PR #1, opened 2019-01-03
Development

adriana-alter:


adriana-alter:
1) Sorry about the enormous, monolithic pull request!!!  Future PRs will be of a normal magnitude.

2) In order to see the changed files, you have to specify that you just want to see changes from the most recent commit--here's a direct link: https://github.com/NYPL-Simplified/registry_admin/pull/1/commits/ba7fa36c0318f3d473e89f57cba9c9c10a14a7f1

3) There are no actual changes in the icons folder (which was copied straight from circulation-web), even though those files show up in the list of changed files; ignore them.

4) Should you wish to run the app, you would need to use the configure-front-end branch of library_registry.

aslagle:
For future reference, when you start a new repo it's nice to have the first commit be something like the license and a mostly empty README - that way everything else can be added in PRs and reviewed.

Also since I can't review it - looks like some of the stuff outside `src` isn't applicable for this and should probably be deleted or cleaned up, and there are many dependencies that aren't needed.

adriana-alter:
I started the repo by cloning from circulation-web so that the basic configuration would already be in place (i.e. there was no point at which the repo was just license and README)...but yes, I definitely still should have deleted way more stuff and added things back one at a time!  Next time :) 

I'll clean up the extraneous stuff outside `src`!

aslagle:
What happens if you merge master into here? That might make it easier to review and if it doesn't work you'll probably have to figure it out anyway to merge this PR.

adriana-alter:
Just tried; interestingly (and unhelpfully!), it says that everything is "already up to date."

aslagle:
That’s weird since github is showing a merge conflict

adriana-alter:
Update: I tried switching to `master`, doing `git pull`, and then switching back to `development` and merging `master` in.  This caused most of the files to disappear from `development`.  Less confusing (since I had deleted a bunch of stuff from `master` in hopes of making this easier to review), but even more unhelpful.  (I pulled the stuff back into development; no crisis.  Just the same annoying situation we already had.)

aslagle:
You might need to revert the commit that deletes src on master.

adriana-alter:
Ok--successfully merged!  (Doesn't look like this has made it easier to review, though...)

aslagle:
I think it did :)

Now the src changes are showing up in the diff so we don’t have to look at individual commits.

adriana-alter:
Wonderful!  (Sorry again about the giant PR; I realize I went about this very wrong...)

EdwinGuzman:
That is weird because I also tried pulling from master and merging into development (and other combinations) and nothing resulted in changes going from one branch into master. Glad it was solved.

package.json
```diff
@@ -21,26 +21,15 @@
     "prod": "rm -rf dist && NODE_ENV=production webpack -p --progress --display-modules --config webpack.prod.config"
   },
   "dependencies": {
-    "@nypl/dgx-svg-icons": "0.3.4",
```
EdwinGuzman: You should use this instead of glyphicon unless if glyphicon renders SVG. Font icons are not accessible and we try not to use them.

adriana-alter: I'm not seeing anything in the NYPL icons file that would make sense to use here.  

Suggested longer-term goal: add an up wedge icon to the NYPL icons thing.  (There are currently down, left, and right wedges, but no up.)

In the meantime, these look nice to me; what do you think? https://glyph.smarticons.co/

adriana-alter: P.S. In the meantime, there are a bunch of font-awesome icons (and two Glyphicons) currently living in circ-web.  Worth making a ticket about...yes?

aslagle:
All the selenium-related stuff can also go - tests/browser, nightwatch and selenium dependencies, commands in package.json. We may want something like that eventually but they don't even work in circulation-web, so no reason to let them proliferate.

package.json
```diff
@@ -21,26 +21,15 @@
...
     "bootstrap": "^3.3.6",
-    "font-awesome": "^4.6.3",
-    "isomorphic-fetch": "^2.2.1",
-    "numeral": "^2.0.6",
-    "opds-feed-parser": "0.0.14",
     "opds-web-client": "0.1.27",
```
aslagle: I'm not sure if it's worth having this as a dependency but I haven't looked at what all it's used for yet.

adriana-alter: The DataFetcher, RequestError, and BaseActionCreator are being imported from it.

aslagle: It kind of seems like overkill to depend on the whole app just for those. If they're that useful maybe they should eventually become a separate package.

adriana-alter: I think they should definitely become a separate package!  But that's a conversation for another branch; let's leave it as is for right now.  (There's no risk of it hurting anything in the meantime...right?)

src/components/LibrariesList.tsx
```diff
@@ -0,0 +1,89 @@
...
+    this.select = this.select.bind(this);
+  }
+
+  isActive(library: LibraryData) {
+    return this.props.libraries.libraries.indexOf(library) === parseInt(this.state.activeKey);
+  }
+
+  select(idx: string) {
+    this.setState({ activeKey: idx});
+    this.props.fetchData();
```
aslagle: Why fetch all the libraries again here?

package.json
aslagle: Yeah that's fine.

src/components/LibrariesList.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  render(): JSX.Element {
+    return(
+      <PanelGroup accordion activeKey={this.state.activeKey}>
+        { this.props.libraries &&
+          this.props.libraries.libraries.map(library =>
+            <LibrariesListItem
+              key={library.uuid}
+              idx={`${this.props.libraries.libraries.indexOf(library)}`}
```
aslagle: Does PanelGroup require this for the activeKey?

src/components/LibrariesList.tsx
adriana-alter: So that, if you edited the stages, the color-coding immediately changes to reflect that.

src/components/LibrariesList.tsx
adriana-alter: Exactly!

src/components/LibrariesList.tsx
```diff
@@ -0,0 +1,89 @@
...
+              library={library}
+              active={this.isActive(library)}
+              select={this.select}
```
aslagle: How does the index argument work?

package.json
```diff
@@ -15,32 +15,19 @@
     "test": "npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js lib/__tests__/*.js lib/**/__tests__/*.js lib/**/**/__tests__/*.js",
     "test-file": "npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js",
     "test-browser": "npm run test-chrome && npm run test-firefox",
```
aslagle: This one should also be removed.

adriana-alter: Which one--`test-browser`?

aslagle: Yes - it calls the ones you deleted already.

src/components/LibrariesListContainer.tsx
```diff
@@ -0,0 +1,46 @@
...
+export interface LibrariesListContainerDispatchProps<T> {
+  fetchData?: () => Promise<T>;
+}
+
+export interface LibrariesListContainerOwnProps {
+  store?: Store<State>;
+  csrfToken?: string;
+}
+
+export interface LibrariesListContainerProps<T> extends LibrariesListContainerDispatchProps<T>, LibrariesListContainerOwnProps {}
```
aslagle: Since this isn't a connected component I'm confused about what these props are doing. Do you even need them?

src/components/LibrariesList.tsx
adriana-alter: It's what tells the panel whether it should be open or closed; if panel's idx matches parent PanelGroup's activeKey, panel is open.

aslagle: But how does `idx` end up being passed into it since it's not on this line? Is something putting it in automatically?

adriana-alter: What do you mean about it not being on this line...?

aslagle: I would expect this line to be `select={() => this.select(idx)}` - I don't know how `select` is getting the index.

src/components/LibrariesListContainer.tsx
adriana-alter: That...is a very good point.  I will get rid of them :)

src/components/LibrariesList.tsx
aslagle: Oh I see, the `ListItem` component is doing it. Sorry I hadn't gotten to that part yet. I think it would be cleaner if the `ListItem` didn't need to know where it was in the list.

src/components/LibrariesList.tsx
adriana-alter: I'll look at it again, but I'm pretty sure...

src/components/LibrariesList.tsx
```diff
@@ -0,0 +1,89 @@
...
+  select(idx: string) {
+    this.setState({ activeKey: idx});
+    this.props.fetchData();
+  }
+
+  render(): JSX.Element {
+    return(
+      <PanelGroup accordion activeKey={this.state.activeKey}>
+        { this.props.libraries &&
+          this.props.libraries.libraries.map(library =>
```
aslagle: I think you can change this to `.map((library, idx) =>` instead of having to compute the index for every library, which could be slow if there are a lot.

src/components/LibraryDetailContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { LibraryData } from "../interfaces";
+import { connect } from "react-redux";
+import ActionCreator from "../actions";
+
+export interface LibraryDetailContainerContext {
+  store: Store<State>;
+}
+
+export interface LibraryDetailContainerOwnProps {
+  id: string;
```
aslagle: For consistency why not name this prop `uuid` like the other components?

src/components/LibraryDetailContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+  toggle: () => void;
+}
+
+export interface LibraryDetailContainerDispatchProps {
+  fetchData?: () => Promise<LibraryData>;
+}
```
aslagle: The dispatch props are also unused here.

src/components/LibraryDetailContainer.tsx
adriana-alter: Will do!

src/components/LibraryDetailPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+}
+
+function mapStateToProps(state: State, ownProps: LibraryDetailPageOwnProps) {
+  return {
+    fullLibrary: state.library && state.library.data
+  };
+}
+
+function mapDispatchToProps(dispatch: Function, ownProps: LibraryDetailPageOwnProps) {
+  let actions = new ActionCreator(null, null);
```
aslagle: If you wanted to make the CSRF token work this would be the main place it's needed. It's not being used on the server yet either.

package.json
EdwinGuzman: I think there's a ticket already to update icons to use svg icons. That SVG icon set does look nice but then you're bringing in yet another set of icons. Also, you can use CSS to rotate the up wedge icon.

package.json
EdwinGuzman: Don't forget to remove later!

package.json
aslagle: I'm not sure if anything from the src/components/icons directory in here is being used. Would it make sense to remove that too?

src/interfaces.ts
```diff
@@ -0,0 +1,40 @@
...
+  label: string;
+  required?: boolean;
+}
+
+export interface LibrariesData {
+  libraries: LibraryData[];
+  settings?: LibrarySettingField[];
+}
+
+export interface LibraryWithSettingsData {
```
aslagle: Doesn't look like you need this or SettingData.

src/components/LibraryDetailPage.tsx
adriana-alter: Do we actually need the CSRF token in the first place (given the very small user base)...?

aslagle: Probably not for now.

src/stylesheets/app.scss
```diff
@@ -0,0 +1,11 @@
+@import "../../node_modules/bootstrap/dist/css/bootstrap.css";
+@import "../../node_modules/font-awesome/css/font-awesome.min.css";
```
aslagle: Did you end up using anything from font awesome?

package.json
EdwinGuzman: If it's not, then yes remove. But if we will use icons, then we should try to use these.

adriana-alter: The ones that were in that directory !== the ones we'd be using for this.  (Or is the solution to move more icons into that directory?)

src/components/LibrariesList.tsx
adriana-alter: Update: it worked!!

src/components/LibrariesList.tsx
adriana-alter: I looked at it again.  Yes, you need both `activeKey` and `expanded`.  Annoying, but not hurting anything.  (And it's much less annoying now that the panel doesn't need the `idx` prop!)

package.json
```diff
@@ -14,33 +14,20 @@
...
-    "test-firefox": "node_modules/.bin/selenium-standalone install && nightwatch -e firefox",
     "dev": "rm -rf dist && NODE_ENV=development webpack --watch --progress --display-modules --config webpack.prod.config",
     "prod": "rm -rf dist && NODE_ENV=production webpack -p --progress --display-modules --config webpack.prod.config"
   },
   "dependencies": {
-    "@nypl/dgx-svg-icons": "0.3.4",
+    "@nypl/dgx-svg-icons": "^0.3.9",
     "bootstrap": "^3.3.6",
-    "font-awesome": "^4.6.3",
-    "isomorphic-fetch": "^2.2.1",
```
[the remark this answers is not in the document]
adriana-alter: No.  Should it be...?

src/actions.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  editStages(data: FormData) {
+    let url = "/admin/libraries/registration";
+    return this.postForm(ActionCreator.EDIT_STAGES, url, data).bind(this);
+  }
+
+
+
+
```
EdwinGuzman: Why all this empty white space?

package.json
```diff
@@ -14,33 +14,20 @@
...
     "react-bootstrap": "0.29.4",
-    "react-color": "^2.14.1",
     "react-dom": "^15.4.0",
     "react-redux": "^5.0.6",
     "react-router": "^3.2.0",
-    "recharts": "^1.0.1",
     "redux": "^3.7.2",
-    "redux-thunk": "^2.1.0",
-    "request": "^2.88.0",
-    "uswds": "^1.6.9"
```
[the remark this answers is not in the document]
adriana-alter: It came from you; you said, let's move towards using the U.S. Web Design System!  (Which I didn't do; I bootstrapped everything for the sake of expediency and consistency.  But at some point...)

src/actions.ts
```diff
@@ -0,0 +1,173 @@
...
+
+
+
+
+
+  postForm(
```
EdwinGuzman: I'd prefer if these functions were declared before their invocations in the three functions you have above. Javascript does the hoisting already but better to be explicit.

src/components/App.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
+import Header from "./Header";
+import LibrariesListContainer from "./LibrariesListContainer";
+
+export default class App extends React.Component<void, void> {
+
+  render(): JSX.Element {
+    return(
+      <div id="main-app-component">
+          <Header />
```
EdwinGuzman: You have an extra indentation here it seems.

src/components/Header.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from "react";
+
+export default class Header extends React.Component<void, void> {
+
+  render(): JSX.Element {
+    return(
+      <nav className="navbar navbar-default">
+        <span className="navbar-brand">Welcome to the library registry interface!</span>
```
EdwinGuzman: "library registry" should probably be capitalized.

src/actions.ts
adriana-alter: As in...1) create an interface for ActionCreator, 2) literally just move the code, or 3) declare them but not define them (i.e. `let postForm;`), which TypeScript finds upsetting?

src/components/LibrariesList.tsx
EdwinGuzman: I agree, `.map` would be better. Also, use `index` instead of `idx` to be clearer.

src/components/LibrariesListContainer.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
+import LibrariesList from "./LibrariesList";
+import { Store } from "redux";
+import { State } from "../reducers/index";
+import { LibrariesData, LibraryData } from "../interfaces";
+import { connect } from "react-redux";
+import ActionCreator from "../actions";
+
+export interface LibrariesListContainerContext {
+  store: Store<State>;
```
EdwinGuzman: @aslagle There was a store issue where a component could not readily connect to the store. I had this issue before and I told Adriana to create this top level component to use the store and pass it down so that the child component can connect to the store. I think we can eventually do without this component since the store in this app only has one value. We'll revisit this issue later (not sure if it actually is an issue though).

src/components/LibrariesListContainer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  context: LibrariesListContainerContext;
+  static contextTypes: React.ValidationMap<LibrariesListContainerContext> = {
+      store: React.PropTypes.object.isRequired,
+      csrfToken: React.PropTypes.string.isRequired,
+  };
+
+  render(): JSX.Element {
+    return(
+      <div className="libraries-list-container">
+        <div>
```
EdwinGuzman: Unnecessary `<div>` here.

src/components/LibrariesListItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export default class LibrariesListItem extends React.Component<LibrariesListItemProps, LibrariesListItemState> {
+  constructor(props: LibrariesListItemProps) {
+    super(props);
+    this.colorCode = this.colorCode.bind(this);
+  }
+
+  header() {
+    return (
+      <div onClick={this.props.select}>
```
EdwinGuzman: `<div>`s should not have an `onClick` attribute unless if you give it a `role="button"` attribute. In that case, you might as well use a `<button>`.

src/components/LibrariesListContainer.tsx
adriana-alter: A lot of online tutorials suggest using this pattern (for whatever that's worth...).

src/components/LibrariesListItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+  colorCode() {
+    // If both stages are in production, background is green;
+    // if at least one stage is cancelled, background is red;
+    // otherwise, background is yellow.
+    let stages = [this.props.library.registry_stage, this.props.library.library_stage];
+
+    if (stages.every((stage) => stage === "production")) {
+      return "success";
+    }
+    else if (stages.some((stage) => stage === "cancelled")) {
```
EdwinGuzman: `else if`s should start on the same line as the ending `}`.

EdwinGuzman: Is `cancelled` a new stage value for the library registry? I don't remember encountering that before for the circulation admin when working on it.

src/components/LibrariesListContainer.tsx
aslagle: Someone should review the latest redux best practices. When we started the recommendation was to connect top-level components to the store and pass everything down to other components in props, but I think at some point they changed to recommend that you connect individual components that need something from the store. That was a while ago too so I don't know what's suggested now.

src/components/LibrariesListItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+    return (
+      <div onClick={this.props.select}>
+        <span>{this.props.library.name} ({this.props.library.short_name})</span>
+        <GenericWedgeIcon className={this.props.active ? "up-icon" : "down-icon"} />
+      </div>
+    );
+  }
+
+  colorCode() {
+    // If both stages are in production, background is green;
```
EdwinGuzman: "If both Library and Registry stages ..."

src/components/LibrariesListItem.tsx
adriana-alter: I don't know how new it is (but it's definitely supposed to be there!)...

src/components/LibrariesListItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+  select: (event: __React.MouseEvent) => void;
+}
+export interface LibrariesListItemState {
+  open: boolean;
+}
+
+export default class LibrariesListItem extends React.Component<LibrariesListItemProps, LibrariesListItemState> {
+  constructor(props: LibrariesListItemProps) {
+    super(props);
+    this.colorCode = this.colorCode.bind(this);
```
EdwinGuzman: It doesn't seem like a big deal but you should bind `header` here, too.

src/components/LibraryStageItem.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export interface LibraryStageItemProps {
+  label: string;
+  value: string;
+}
+export default class LibraryStageItem extends React.Component<LibraryStageItemProps, void> {
+  render(): JSX.Element {
+    return(
+      <fieldset className="form-group">
+        <label>{this.props.label}: {this.props.value}</label>
```
EdwinGuzman: This should be `<legend>` instead of `<label>` for `<fieldset>`s.

package.json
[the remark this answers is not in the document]
EdwinGuzman: I'm just confused why it wasn't but not a big deal.

package.json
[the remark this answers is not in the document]
EdwinGuzman: Yes, at some point. I said we'd be using something like it but creating our own, not that you had to use this specific package now!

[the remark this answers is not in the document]
adriana-alter: Okay, okay--I'll get rid of it!

src/actions.ts
EdwinGuzman: 2)

src/components/LibrariesListContainer.tsx
EdwinGuzman: I can pick it up because I don't have too much to work on now.

src/components/LibrariesListItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+    // otherwise, background is yellow.
+    let stages = [this.props.library.registry_stage, this.props.library.library_stage];
+
+    if (stages.every((stage) => stage === "production")) {
+      return "success";
+    }
+    else if (stages.some((stage) => stage === "cancelled")) {
+      return "danger";
+    }
+    else {
```
EdwinGuzman: You don't need this last `else` statement.

src/components/LibrariesListItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { LibraryData } from "../interfaces";
+import LibraryDetailContainer from "./LibraryDetailContainer";
+import { GenericWedgeIcon } from "@nypl/dgx-svg-icons";
+
+export interface LibrariesListItemProps {
+  library: LibraryData;
+  active: boolean;
+  select: (event: __React.MouseEvent) => void;
+}
+export interface LibrariesListItemState {
```
EdwinGuzman: You are not actually using the `open` state value so you do not need this.

src/components/LibrariesListItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+    let style = this.colorCode();
+
+    return(
+      <Panel
+        bsStyle={style}
+        header={this.header()}
+        collapsible={true}
+        expanded={this.props.active}
+      >
+        { this.props.active &&
```
EdwinGuzman: Having or not having `this.props.active` makes no difference for rendering/showing `LibraryDetailsContainer`.

src/components/LibrariesList.tsx
EdwinGuzman: The problem with this is that it's fetching all libraries data every time you open and close a panel. It's unnecessary if an admin wants to just look at all the library data and not make any changes.

Also, I noticed there was a flash every time I opened or closed a panel and that it wasn't smooth like the examples in the react-bootstrap docs. This fetch to the API causes that flash and non-smooth transition.

EdwinGuzman: This seems to be the component tree:
```
  LibrariesList
    LibrariesListItem
      LibraryDetailContainer
        LibraryDetailPage
```

From `LibraryDetailPage` when you submit you then trigger a toggle/select function which triggers all the way up to this `select` function. It's a long way for the panel just to close after submitting data.

You can probably pass a callback from `ListDetailPage` with the value of the Library and Registry stages up to `LibrariesListItem`. There you can intercept the `this.props.select` call, get the two stage values, and render the panel header with the appropriate color. Then you would not have to fetch all libraries data again. It also seems you don't need all that information anyway because when `LibraryDetailPage` renders, it fetches its own set of data.

src/components/LibraryDetailContainer.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  context: LibraryDetailContainerContext;
+  static contextTypes: React.ValidationMap<LibraryDetailContainerContext> = {
+      store: React.PropTypes.object.isRequired,
+  };
+
+  render(): JSX.Element {
+    return(
+      <div id="library-detail-container">
+        <LibraryDetailPage store={this.context.store} uuid={this.props.uuid} toggle={this.props.toggle}/>
```
EdwinGuzman: Space before the `/>`. Also, is eslint not picking these issues up?

src/components/LibrariesList.tsx
adriana-alter: Ah, I wondered why there was a flash!  I will give this a try.

src/components/LibraryDetailContainer.tsx
```diff
@@ -0,0 +1,35 @@
...
+export default class LibraryDetailContainer extends React.Component<LibraryDetailContainerProps, void> {
+
+  context: LibraryDetailContainerContext;
+  static contextTypes: React.ValidationMap<LibraryDetailContainerContext> = {
+      store: React.PropTypes.object.isRequired,
+  };
+
+  render(): JSX.Element {
+    return(
+      <div id="library-detail-container">
```
EdwinGuzman: The top level element in `LibraryDetailPage` renders a `div`, so why not remove this one here and move the class name to the `LibraryDetailPage`?

src/components/LibraryDetailContainer.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
```
EdwinGuzman: Do you need this component because you need to pass the store to call it for the child component? I think that was only the top level case but any child should work, no?

src/components/LibrariesListItem.tsx
EdwinGuzman: Now I'm wondering what exactly this `active` prop does?

src/components/LibraryDetailContainer.tsx
adriana-alter: That's why I have it, but I'll try getting rid of it and see what happens...

src/components/LibrariesListItem.tsx
adriana-alter: It 1) sets the value for `expanded` (which the panel needs, because Bootstrap), and 2) determines whether the icon is the up arrow or the down arrow.

src/components/LibraryDetailItem.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+
+export interface LibraryDetailItemProps {
+  label: string;
+  value: string;
+}
+export default class LibraryDetailItem extends React.Component<LibraryDetailItemProps, void> {
+  render(): JSX.Element {
+    return(
+      <section className="list-group-item">
```
EdwinGuzman: `section`s should be used for major sections on the page, like the header, the left/right sidebar, or the main container. Here it doesn't make too much sense to use for every library detail. Since this is a list, you should use the `li`. In `LibraryDetailPage`, the div with class `list-group` should be a `ul`.

src/components/LibraryDetailPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  render(): JSX.Element {
+    if (!this.props.fullLibrary) {
+      return <div></div>;
+    }
+    return(
+      <div>
+        { this.renderStages() }
+        <div id="info" className="list-group">
```
EdwinGuzman: This should be a `ul`.

src/components/LibraryDetailPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+    let form = (this.refs["form"] as any);
+    const data = new (window as any).FormData(form);
+    await this.props.editStages(data);
+    this.props.toggle(event);
+    this.props.fetchData();
+  }
+
+  render(): JSX.Element {
+    if (!this.props.fullLibrary) {
+      return <div></div>;
```
EdwinGuzman: Rendering null would be better than an empty div.

src/components/LibraryDetailItem.tsx
EdwinGuzman: You could also structure this as a definition list (`dl`) from the label/p structure you defined before, and that would be fine too.

src/components/LibraryDetailPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <LibraryStageItem key={"libraryStage"} label={"Library Stage"} value={libraryStage} />
+        <LibraryStageItem key={"registryStage"} label={"Registry Stage"} value={registryStage} />
+        <SaveButton
+          submit={this.submit}
+        />
+      </form>
+    );
+  }
+
+async submit(event: __React.MouseEvent): Promise<void> {
```
EdwinGuzman: This line should be indented.

src/components/LibraryDetailPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+    let libraryStage = this.props.fullLibrary["library_stage"];
+    let registryStage = this.props.fullLibrary["registry_stage"];
+    return (
+      <form ref="form" className="well">
+        <input
+          type="hidden"
+          name="uuid"
+          value={this.props.uuid}
+        />
+        <LibraryStageItem key={"libraryStage"} label={"Library Stage"} value={libraryStage} />
```
EdwinGuzman: You don't need the `key` attribute here.

src/components/LibraryDetailPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+      </form>
+    );
+  }
+
+async submit(event: __React.MouseEvent): Promise<void> {
+    event.preventDefault();
+    let form = (this.refs["form"] as any);
+    const data = new (window as any).FormData(form);
+    await this.props.editStages(data);
+    this.props.toggle(event);
```
EdwinGuzman: Won't this be annoying that once you update a library the panel closes? Why not leave it open and let the user decide if they want to close it?

src/components/LibraryDetailPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+    this.props.fetchData();
```
EdwinGuzman: This seems unnecessary because you are fetching data after submitting but before fetching the data you are closing the panel. When you close the panel you unmount the component so the new data won't be displayed anyway. You're relying on the top level `libraries` data for the stage level for the heading color so another reason why this doesn't seem necessary. When you open the panel and mount the component again, then you do fetch and render the latest data. But, this last part about mounting/unmounting also depends in `LibrariesListItem` where you check `this.props.active` to render the `ListDetailContainer`.

src/components/LibraryDetailPage.tsx
adriana-alter: I thought it would be more annoying for the user to have to close it (since the stages are the only thing you can update; it's not like there's anything else you can do afterwards).  But I'll try it the other way and we'll see!

src/components/LibraryDetailPage.tsx
EdwinGuzman: Also, what you do decide to do here depends on refactoring so that the flash doesn't happen when a panel opens and closes.

src/components/LibraryDetailPage.tsx
EdwinGuzman: If you go with definition list, this would be a `dl`.

src/components/LibraryDetailPage.tsx
adriana-alter: It is completely unnecessary; good catch!

EdwinGuzman:
I don't like react-bootstrap's DOM structure for the `PanelGroup` and `Panel` components because they're just divs and could be more semantic for accessibility. The flash that happens now when opening or closing a panel also makes the page lose focus when you are tabbing through.

src/components/LibrariesListContainer.tsx
```diff
@@ -0,0 +1,29 @@
...
+export default class LibrariesListContainer extends React.Component<void, void> {
+  context: LibrariesListContainerContext;
+  static contextTypes: React.ValidationMap<LibrariesListContainerContext> = {
+      store: React.PropTypes.object.isRequired
+      // csrfToken: React.PropTypes.string.isRequired,
+  };
+
+  render(): JSX.Element {
+    return(
+      <div className="libraries-list-container">
```
EdwinGuzman: This div doesn't seem necessary.

src/components/__tests__/App-test.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <App />
+    );
+  });
+  it("should render the header", () => {
+    let header = wrapper.find(Header);
+    expect(header.length).to.equal(1);
+  });
+  it("should render the libraries list container", () => {
+    let container = wrapper.find(LibrariesListContainer);
+    expect(container.length).to.equal(1);
```
EdwinGuzman: Would you expect this to have a length one if no data is passed to it on a shallow mount?

src/components/__tests__/ContextProvider-test.tsx
```diff
@@ -0,0 +1,34 @@
+import { expect } from "chai";
+import * as Enzyme from "enzyme";
+
+import * as React from "react";
+import { shallow } from "enzyme";
```
EdwinGuzman: You should be able to combine this and line 2, no?

src/components/__tests__/LibrariesList-test.tsx
```diff
@@ -0,0 +1,61 @@
+import { expect } from "chai";
+import { stub } from "sinon";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+import { shallow, mount } from "enzyme";
```
EdwinGuzman: You're not using `shallow` here or in a few other test files.

src/components/__tests__/LibrariesList-test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    });
+
+    it("should render each library's  name and short name", () => {
+      let lib1 = wrapper.find(".panel").at(0);
+      expect(lib1.text()).to.contain("Test Library 1 (lib1)");
+      let lib2 = wrapper.find(".panel").at(1);
+      expect(lib2.text()).to.contain("Test Library 2 (lib2)");
+    });
+
+    it("should initially have no active key", () => {
```
EdwinGuzman: You should have a test that triggers the select function and test a different state value.

adriana-alter:
@EdwinGuzman Re: react-bootstrap's un-semantic DOM structure for Panel and PanelGroup--I will see what can be done, but before I start messing with it: what do you think the correct DOM structure would be?  `div` for PanelGroup, `section` for Panel?  Or `ul`/`li`?  Or...?

(I don't think the flash and the tabbing issue are react-bootstrap's fault--I think it's because I have the API calls in a weird place.  We'll see.)

src/components/LibraryDetailContainer.tsx
adriana-alter: You called it; component is not needed :)

src/components/__tests__/App-test.tsx
adriana-alter: The container would still be there--it would just be empty...

EdwinGuzman: Is that a good test though?

EdwinGuzman: Or is that what a React component should return when no data is passed to it?

EdwinGuzman:
It's a list of libraries so it should be `ul`/`li` and that whole list can be wrapped in a `section`.

src/components/__tests__/App-test.tsx
adriana-alter: I think the point is that the top-level component for the whole app should still render the same basic architecture irrespective of whether there happens to be data.  Initializing the app with no data, e.g. if there are no libraries in this registry, should always give you a container (it doesn't matter whether the container is empty or full; the container doesn't even know anything about that)--it shouldn't give you a blank screen or an error message or anything like that.

EdwinGuzman: A fallback component (or error message) to render when there is no data to display is a better approach instead of displaying nothing. Then the user would be aware of what is going on instead of getting no feedback. Granted, this app won't have a big user base, but it's good to get into the mindset of building more progressive apps regardless of who uses it for the sake of good implementation habits.

src/reducers/createFetchEditReducer.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export interface FullAction extends Action {
+    error?: RequestError | String;
+    data?: Object;
+    url?: String;
+}
+
+export interface FetchEditReducer<T> {
+  (state: FetchEditState<T>, action: FullAction): FetchEditState<T>;
```
EdwinGuzman: This `FullAction` is nice, better than `any`.

src/stylesheets/mixins.scss
```diff
@@ -0,0 +1,11 @@
+/// Slightly lighten a color
+/// @access public
+/// @param {Color} $color - color to tint
+/// @param {Number} $percentage - percentage of `$color` in returned color
+/// @return {Color}
+
+$white: #FFF !default;
```
EdwinGuzman: You can define this in `colors.scss` and rearrange the order of `mixins` and `colors` in the main scss file.

src/stylesheets/welcome_page.scss
```diff
@@ -0,0 +1,6 @@
+nav.navbar-default span.navbar-brand {
+  color: white;
```
EdwinGuzman: Use variables. Also, the `span.navbar-brand` could be nested.

adriana-alter:
Update: we now have a reusable `Panel` component (not involving react-bootstrap).  No flash; no tabbing issue; no weird clunky combinations of props.

src/stylesheets/mixins.scss
adriana-alter: Turns out I can't; `colors` uses the `tint` method (which uses `$white`) from `mixins`...so if `mixins` needs to import `$white` from `colors`, we've got an endless loop...

src/components/Form.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return(
+      <form ref="form">
+        <input
+          type="hidden"
+          name={this.props.hiddenName}
+          value={this.props.hiddenValue}
+        />
+        { this.props.content }
+        <SaveButton
+          submit={this.save}
```
EdwinGuzman: Would `submit` be a better name than `save` since you don't always save data when you submit a form?

src/components/Header.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from "react";
+
+export default class Header extends React.Component<void, void> {
+
+  render(): JSX.Element {
+    return(
+      <nav className="navbar navbar-default">
+        <span className="navbar-brand">Welcome to the Library Registry interface!</span>
```
EdwinGuzman: Semantically, some message like this should not be in the `nav` bar but could be an `h1` on the page. You'd have the main name of the app in the header/nav but not necessarily a welcome message there.

src/components/Form.tsx
adriana-alter: Yes.  Will do.

src/components/Header.tsx
adriana-alter: It's just a placeholder; I figure the point of the header is that there will eventually be a logout button there.  (I'll change it anyway, though; might as well!)

src/components/LibrariesList.tsx
```diff
@@ -0,0 +1,68 @@
...
+export interface LibrariesListDispatchProps {
+  fetchData: () => Promise<LibrariesData>;
+}
+
+export interface LibrariesListProps extends LibrariesListStateProps, LibrariesListOwnProps, LibrariesListDispatchProps {};
+
+export class LibrariesList extends React.Component<LibrariesListProps, void> {
+
+  render(): JSX.Element {
+    let hasLibraries = (this.props.libraries && this.props.libraries.libraries && this.props.libraries.libraries.length);
```
EdwinGuzman: Not a big deal but this `has...` name sounds more like a boolean but it's actually a numerical value.

src/components/LibrariesList.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <ul className="list panel-group">
+        { hasLibraries ?
+          this.props.libraries.libraries.map(library =>
+            <LibrariesListItem
+              key={library.uuid}
+              library={library}
+              store={this.props.store}
+            />
+          ) :
+          <h1 className="page-header">There are no libraries in this registry yet.</h1>
```
EdwinGuzman: The welcome message, or rather, the name of the app, should be an h1. This is more like an h2 or maybe not even a heading at all since it's not defining something on the page.

src/components/LibrariesList.tsx
adriana-alter: That...is because it was supposed to be a boolean but I wrote it wrong.  Will fix.

src/components/LibrariesListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+          library={this.props.library}
+          updateColor={this.updateColor}
+          store={this.props.store}
+        />
+    );
+  }
+
+  updateColor(library_stage: string, registry_stage: string): void {
+      let color = this.colorCode(library_stage, registry_stage);
+      this.setState({ color: color });
```
EdwinGuzman: You can use the short hand `this.setState({ color });` if you want to.

src/components/LibraryDetailPage.tsx
```diff
@@ -0,0 +1,105 @@
...
+          <LibraryStageItem key="lib" label={"Library Stage"} value={this.state.libraryStage} />,
+          <LibraryStageItem key="reg" label={"Registry Stage"} value={this.state.registryStage} />
+        ]}
+      />
+    );
+  }
+
+  async submit(data): Promise<void> {
+    this.props.updateColor(data.get("Library Stage"), data.get("Registry Stage"));
+    await this.props.editStages(data);
```
EdwinGuzman: It's possible this could fail, no? If it does, you'd be updating the color to the right value but because the server couldn't update the values, they wouldn't be right.

src/components/LibraryStageItem.tsx
```diff
@@ -0,0 +1,29 @@
...
+      "testing": "warning",
+      "production": "success",
+      "cancelled": "danger"
+    };
+
+    return(
+      <fieldset className="form-group well">
+        <legend>
+          <span>{this.props.label}</span>
+          <label className={`label label-${colors[this.props.value]}`}>{this.props.value}</label>
```
EdwinGuzman: This may be too long but something like `current library stage: ` would be better.

src/components/__tests__/LibraryDetailPage-test.tsx
```diff
@@ -0,0 +1,83 @@
+import { expect } from "chai";
+import { stub, spy } from "sinon";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+import { mount } from "enzyme";
+import buildStore from "../../store";
+import LibraryDetailPage from "../LibraryDetailPage";
+
+describe("LibraryDetailPage", () => {
```
EdwinGuzman: This component should have a behavior test for the `submit` function.

src/stylesheets/colors.scss
```diff
@@ -0,0 +1,23 @@
+$green-tint: #F3F7E6;
+$green-bright: #008918;
+$medium-gray: #DDD;
+
+// Some NYPL specific colors: https://nypl.github.io/design-toolkit/sections/color.html.
+$nypl-red: #D0343A;
```
EdwinGuzman: We are using NYPL specific colors but for circulation manager we didn't use the `$nypl-` prefix because the project can be run by anyone. The same idea applies here since NYPL and DPLA (and anyone else) can use this registry admin.

src/stylesheets/form.scss
```diff
@@ -0,0 +1,48 @@
...
+    border-radius: .25em;
+    font-weight: bold;
+    transition: background 1s, color 1s;
+    &:hover{
+      background: $nypl-light-blue;
+      color: $nypl-dark-gray;
+    }
+  }
+
+  fieldset label.label-success {
```
EdwinGuzman: Group in SCSS style so
```
fieldset {
  label {
   &.label-success { ...}
   &.label- ....
   ...
  }
}
```

EdwinGuzman:
Looks good but
- the orange and green (with some exceptions) background with white foreground for the hover state on the panel heading is not accessible.
- the yellow background/white foreground stage label is not accessible.
- a panel doesn't close when a new one opens. I'm not sure if you still wanted that functionality in place.
- a more explicit explanation for the current stage label would be better to not have users (I know it's a low user base) guess what it means.

EdwinGuzman:
https://nypl.github.io/design-toolkit/sections/color-a11y.html is useful to see how color combinations fare for color contrast.

screen shot 2019-01-10 at 2 46 52 pm

src/components/LibraryDetailPage.tsx
adriana-alter: Yes.  But I haven't done anything with error handling anywhere, yet; I figure that's in the next phase...

EdwinGuzman:
I think your PR is at a stable point where @aslagle and I can review again. You can keep doing more refactoring for the next PR.

EdwinGuzman:
Looks good. Are you still going to work on closing the current panel when another panel is open?

adriana-alter:
I am leaving the closing-the-current-panel thing alone.  At least for now.  (Aside from anything else: I had forgotten that there is actually going to be a second interactive thing inside the panel, so that you might want to edit the stages AND do something else...so, as you predicted, it would be potentially annoying for the panel to close itself.)

EdwinGuzman:
One thing that should be changed, as discussed before, the H1 in the nav should not be there.

adriana-alter:
Well, you approved the PR before I pushed the latest set of changes!  The H1 in the nav is about to not be there.  (But might as well do that in the next PR at this point.)

src/components/ContextProvider.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+import { Store } from "redux";
+import buildStore from "../store";
+import { State } from "../reducers/index";
+
+export interface ContextProviderProps extends React.Props<any> {
+  csrfToken: string;
```
aslagle: is it okay for this to still be a required prop when you're not passing it down from the server? i see you commented out the other uses in this file.

src/components/LibrariesListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+import LibraryDetailPage from "./LibraryDetailPage";
+import { GenericWedgeIcon } from "@nypl/dgx-svg-icons";
+
+export interface LibrariesListItemProps {
+  library: LibraryData;
+  store: Store<State>;
+}
+
+export interface LibrariesListItemState {
+  color: string;
```
aslagle: I'm not sure why this needs to be in the state. Is it so the color can update before the request to change it on the server?

src/components/ContextProvider.tsx
adriana-alter: It doesn't seem to be breaking anything.  (But in any event: I'm currently working on a branch for implementing login; that will get rid of all the unnecessary CSRF stuff!)

src/components/LibrariesListItem.tsx
adriana-alter: Exactly!

aslagle: This isn't great if the request causes an error - seems like the header color changes but the current stage and its color stay the same

adriana-alter: Agreed...but there's no error handling anywhere yet.  I figured that would be in the next phase.

src/components/reusables/Form.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from "react";
+import SaveButton from "./SaveButton";
+
+export interface FormProps {
+  content: Array<JSX.Element>;
+  onSubmit: any;
+  hiddenName?: string;
+  hiddenValue?: string;
```
aslagle: in theory a form could have multiple hidden inputs. i'm not sure this needs to be separated out specifically, it could be part of the content.

src/components/reusables/Form.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export default class Form extends React.Component<FormProps, void> {
+
+  constructor(props: FormProps) {
+    super(props);
+    this.submit = this.submit.bind(this);
+  }
+
+  submit(event: __React.MouseEvent): any {
```
aslagle: The return type should probably be void?

src/components/reusables/Header.tsx
```diff
@@ -0,0 +1,19 @@
...
+  imgSrc?: string;
+}
+
+export default class Header extends React.Component<HeaderProps, void> {
+  render(): JSX.Element {
+    return(
+      <nav className="navbar navbar-default">
+        <div className="navbar-header">
+          <span>{this.props.text}</span>
+          <img className="navbar-brand img-rounded" alt="SimplyE" src={require(`${this.props.imgSrc}`)} />
```
aslagle: Not sure why this is reusable, but if it is meant to be reused you wouldn't want to hard-code "SimplyE".

src/components/reusables/Panel.tsx
```diff
@@ -0,0 +1,46 @@
...
+  constructor() {
+    super();
+    this.renderHeader = this.renderHeader.bind(this);
+    this.toggle = this.toggle.bind(this);
+    this.state = { display: "collapse", icon: "down-icon" };
+  }
+
+  toggle() {
+    let display = this.state.display === "collapse" ? "" : "collapse";
+    let icon = this.state.icon === "down-icon" ? "up-icon" : "down-icon";
```
aslagle: Why not decide the icon based on state.display so you don't need to keep both in the state?

src/reducers/createFetchEditReducer.ts
```diff
@@ -0,0 +1,120 @@
...
+
+      case `${fetchPrefix}_${ActionCreator.LOAD}`:
+        return Object.assign({}, state, {
+          data: action.data,
+          isFetching: false,
+          isLoaded: true
+        });
+
+      default:
+        if (extraActions) {
```
aslagle: I would leave this out until it's actually needed for something, or move the code from circulation-web to a reusable place.

adriana-alter: You mean take out the `extraActions` stuff...right?

src/components/reusables/Form.tsx
adriana-alter: It will always need to have at least one hidden input, though...right?  I figure any additional hidden inputs can be part of the content...

aslagle: I don't know that it necessarily would, it would depend on the form. Creating a new thing would be less likely to have one than editing an existing thing (that may not apply for the registry though)

adriana-alter: To my knowledge, the registry interface is never going to have a creating-a-new-thing form.  Worth leaving alone for now, and then reassessing if we ever try to reuse this component elsewhere (maybe give it an optional `edit` prop to indicate whether it should have the hidden input)?

src/components/reusables/Header.tsx
```diff
@@ -8,11 +8,12 @@ export interface HeaderProps {
 
 export default class Header extends React.Component<HeaderProps, void> {
   render(): JSX.Element {
+    let src = this.props.imgSrc ? require(`${this.props.imgSrc}`) : "";
     return(
       <nav className="navbar navbar-default">
         <div className="navbar-header">
           <span>{this.props.text}</span>
-          <img className="navbar-brand img-rounded" alt={this.props.alt} src={require(`${this.props.imgSrc}`)} />
+          <img className="navbar-brand img-rounded" alt={this.props.alt} src={src} />
```
EdwinGuzman: You should make `src` above default to undefined and then wrap this in a conditional because you do not want to render an empty image element.

src/components/reusables/__tests__/Header-test.tsx
```diff
@@ -4,7 +4,7 @@ import * as React from "react";
 import * as Enzyme from "enzyme";
 import Header from "../../reusables/Header";
 
-describe("Header", () => {
+describe.only("Header", () => {
```
EdwinGuzman: Remove this `.only`.

src/components/reusables/Form.tsx
EdwinGuzman: That may be right, but I think it's also more of creating generic and reuseable components without depending just on the use case. Then it'll be easier to update and extend later on than to have coupled components that will require more work to extend.

src/components/LibrariesListItem.tsx
adriana-alter: Ok--what do you think of this implementation?  Still using `state` in `LibrariesListItem`, but passing it values from the server.  (It seems like overkill to have the `LibrariesList` fetch all new data just for the sake of updating the header for one library...and also seems like overkill to connect `LibrariesListItem` to the store just to do this...right?)

aslagle: I'm not sure exactly what you mean about passing it values from the server. If the values are coming from the server, wouldn't they be in the props already?

One thing you could consider is to update the redux store after the edit request succeeds - we haven't used this pattern, but you can have an action that updates something in the store without making a request to the server. Maybe that would be cleaner?

src/components/LibraryDetailPage.tsx
```diff
@@ -0,0 +1,112 @@
...
+  constructor(props: LibraryDetailPageProps) {
+    super(props);
+    this.submit = this.submit.bind(this);
+    this.renderInfo = this.renderInfo.bind(this);
+    this.renderStages = this.renderStages.bind(this);
+    this.state = { libraryStage: this.props.library.library_stage, registryStage: this.props.library.registry_stage };
+  }
+
+  renderInfo() {
+    let library = this.props.fullLibrary || this.props.library;
```
EdwinGuzman: This is a bit weird that you're expecting data from the server and if it's not there then you're expecting data that was passed down to it. But that data was originally from the server at a higher level. What's the difference and why did you put this in?

adriana-alter: Ok, I was really tired when I wrote this yesterday, so it's entirely possible that as soon as I look at it this morning I'll realize that there was a much better way to do it, but as it currently stands:

1) When `LibrariesList` loads, it A) gets a list of LibraryData from the server (i.e. data about every library), and then B) maps through the list and creates the `LibrariesListItems` by passing one LibraryData to each one.  That's `this.props.library`.

2) So when you open a `LibraryDetailPage` (by expanding a panel), the data you see is coming from `this.props.library`.  (Originally, I was just giving the `LibrariesListItems` minimal data, and then having `LibraryDetailPage` look up the rest of the data when you open the panel.  But that was giving us the weird flash.  And unnecessarily complicating the data flow.)

3) When you edit a library from inside `LibraryDetailPage`, assuming there are no errors, `LibraryDetailPage` fetches the newly-updated information about that one library.  This is `this.props.fullLibrary`.  It uses that information to update the panel color and the label colors, as well as the list of the library's properties (i.e., if you edit the library_stage and/or registry_stage, their values should be updated).  

TLDR: when you first open the panel, `LibraryDetailPage` is getting its data from `this.props.library` (passed down from `LibrariesList`).  It doesn't have anything called `this.props.fullLibrary` yet, since it only gets that after you hit the save button.  If you hit save, `LibraryDetailPage` fetches the new data for this library (with the changes you just made) and uses that to re-render.    

I feel like I am probably missing something obvious.  But the only other ways I see to do this would be A) to have the entire `LibrariesList` re-render every time you edit one library (which--assuming that there will eventually be a LOT of libraries--seems like a recipe for performance issues), or B) to connect `LibrariesListItem` to the store and have it re-render when you edit the library...but that's the same pattern as what I currently have, just one level up, so that doesn't seem better, particularly given that `LibraryDetailPage` already has to be connected to the store anyway.

src/components/LibrariesListItem.tsx
adriana-alter: I mean I'm passing it new values from the server: the new data for this library, complete with the changes that the user has made via the form.  So that it re-renders with the new values for library_stage and registry_stage.  (See the long and semi-coherent essay that I wrote in response to Edwin's comment, at the bottom of the page!)

I'm not sure I understand how that pattern would work.  Do you mean, after the edit request succeeds, manually give the store the new library_stage and registry_stage values (so that the data in the store matches the data in the server, but doesn't actually come from the server)...?  At that point, why not just leave the store out of it and update the component's state with those values instead?  (Like what I had before, but waiting until after the edit request has succeeded.)

src/components/LibraryDetailPage.tsx
EdwinGuzman: Hmm interesting but I would say a look for another PR.

src/components/__tests__/LibrariesListItem-test.tsx
```diff
@@ -3,7 +3,6 @@ import { stub } from "sinon";
 
 import * as React from "react";
 import * as Enzyme from "enzyme";
-import { mount } from "enzyme";
```
EdwinGuzman: What's the benefit of importing all instead of just `mount`?

src/components/LibraryDetailPage.tsx
```diff
@@ -43,7 +43,8 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
 
   renderInfo() {
     let library = this.props.fullLibrary || this.props.library;
-    let fields = Object.keys(library).map(label =>
+    // Only create LibraryDetailItems for fields which actually have a value.
```
EdwinGuzman: When is the case when there's a library without a label value?

src/components/__tests__/LibrariesListItem-test.tsx
adriana-alter: For specifying the type for the wrapper--`Enzyme.CommonWrapper`.  This was during the course of getting rid of some TypeScript error messages a while ago.  (On the other hand, there's no benefit to specifically importing `mount` rather than just calling `Enzyme.mount`...)

src/components/LibrariesListItem.tsx
aslagle: This is exactly what I meant - thanks for explaining it better :)

src/components/LibraryDetailPage.tsx
adriana-alter: Optional fields.  For example: a library can, in theory, have a description...but if this library doesn't happen to have a description, then there's no point making a `LibraryDetailItem` that just says "description" with no value.

src/components/LibrariesListItem.tsx
adriana-alter: I agree about closing out this PR and experimenting with implementing this pattern in a different branch, given that the current implementation, while potentially not optimal, does actually work and doesn't seem to be hurting anything.  @aslagle, what do you think?

src/components/LibraryDetailPage.tsx
EdwinGuzman: Ah you know what mixed me up? That you are using the `library` name in `label => library[label]`. If it was `field[label]` it would be easier to know what is going on.

EdwinGuzman: But It makes sense.

src/components/LibrariesListItem.tsx
adriana-alter: @aslagle The more I think about it, I really really want to just close out this horrible giant PR and make further changes in separate branches.  Any strenuous objections?

PR #3, opened 2019-01-22
Tabs

adriana-alter:
Reusable Tabs component to organize library info inside panels.  Goes with the `categories` library_registry branch.
screen shot 2019-01-22 at 5 11 35 pm


EdwinGuzman:
Is there a related JIRA ticket for this?

adriana-alter:
No; I was conceptualizing it as an extension of the work I was doing on the `development` branch, not as a separate task.  I just broke it out into its own branch for convenience, because the `development` PR had gotten so enormous.  (But I'll make a JIRA ticket if you think there should be one anyway!)

EdwinGuzman:
Yeah, there should always be a related JIRA ticket now.

adriana-alter:
Will do!

adriana-alter:
https://jira.nypl.org/browse/SIMPLY-1666

EdwinGuzman:
I'm getting a bunch of errors:

screen shot 2019-01-23 at 11 01 31 am
screen shot 2019-01-23 at 11 01 40 am

I thought it was a session issue so I opened in an incognito window and I got `AttributeError: 'ViewController' object has no attribute 'generate_csrf_token'` (line 155 in /controller.py). I did switch to the new branch for library_registry.

adriana-alter:
Are you using the right server branch?  `categories`

adriana-alter:
There are two new branches for library_registry; you're definitely on `categories`, not `configure-front-end`...right?

EdwinGuzman:
Yes. I read the description for this PR.

src/components/LibrariesListItem.tsx
```diff
@@ -18,7 +18,11 @@ export interface LibrariesListItemState {
 export default class LibrariesListItem extends React.Component<LibrariesListItemProps, LibrariesListItemState> {
   constructor(props: LibrariesListItemProps) {
     super(props);
-    this.state = { color: this.colorCode(this.props.library.library_stage, this.props.library.registry_stage) };
+    this.state = { color: this.colorCode(
```
EdwinGuzman: This would be easier to read if it was:
```
const color = this.colorCode( ....);
this.state = { color };
```

src/components/LibrariesListItem.tsx
```diff
@@ -56,8 +59,9 @@ export default class LibrariesListItem extends React.Component<LibrariesListItem
 
   render(): JSX.Element {
     let style = this.state.color;
-    let headerText = `${this.props.library.name} (${this.props.library.short_name})`;
-
+    let name = this.props.library.basic_info.name;
```
EdwinGuzman: You can use object deconstruction for this.

src/components/LibraryDetailPage.tsx
```diff
@@ -20,13 +21,12 @@ export interface LibraryDetailPageStateProps {
...
   store: Store<State>;
-  uuid: string;
   updateColor: (library_stage: string, registry_stage: string) => void;
 }
 
 export interface LibraryDetailPageState {
-  libraryStage: string;
-  registryStage: string;
+  libraryStage?: string;
+  registryStage?: string;
```
EdwinGuzman: Why are these now optional?

src/components/LibraryDetailPage.tsx
```diff
@@ -36,26 +36,35 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
...
     super(props);
     this.submit = this.submit.bind(this);
-    this.renderInfo = this.renderInfo.bind(this);
+    this.renderItems = this.renderItems.bind(this);
     this.renderStages = this.renderStages.bind(this);
-    this.state = { libraryStage: this.props.library.library_stage, registryStage: this.props.library.registry_stage };
+    this.state = {
+                  libraryStage: this.props.library.stages.library_stage,
+                  registryStage: this.props.library.stages.registry_stage,
+                 };
```
EdwinGuzman: Closing brace should line up with `this....`.

src/components/LibraryDetailPage.tsx
```diff
@@ -65,13 +74,12 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
...
+  async submit(data: FormData): Promise<void> {
     await this.props.editStages(data);
-    await this.props.fetchLibrary(this.props.uuid);
-
-    let libraryStage = this.props.fullLibrary.library_stage;
-    let registryStage = this.props.fullLibrary.registry_stage;
+    await this.props.fetchLibrary(this.props.library.uuid);
 
+    let libraryStage = this.props.fullLibrary.stages.library_stage;
+    let registryStage = this.props.fullLibrary.stages.registry_stage;
```
EdwinGuzman: You should update the interface.

src/components/LibraryDetailPage.tsx
```diff
@@ -83,9 +91,13 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
...
       <div>
         { this.renderStages() }
-        <ul id="info" className="list-group">
-          { this.renderInfo() }
-        </ul>
+        <hr></hr>
+        <div className="detail-content">
+          <Tabs items={{
+            "Basic Information": this.renderItems("basic_info"),
+            "Contact & URLs": this.renderItems("urls_and_contact"),
```
EdwinGuzman: The `renderItems` would be easier to test as a function if you also passed in the library object instead of just the key.

src/components/LibraryStageItem.tsx
```diff
@@ -6,7 +6,8 @@ export interface LibraryStageItemProps {
 }
 export default class LibraryStageItem extends React.Component<LibraryStageItemProps, void> {
   render(): JSX.Element {
-    let colors = {
+    let colors: {[index: string]: string};
```
EdwinGuzman: Why do you need this? There should be a space in between the braces.

src/components/__tests__/LibrariesList-test.tsx
```diff
@@ -10,27 +10,46 @@ describe("LibrariesList", () => {
   let libraries = {
     libraries:
       [{
-        "uuid": "UUID1",
-        "name": "Test Library 1",
-        "short_name": "lib1",
-        "id": 1,
-        "registry_stage": "testing",
-        "library_stage": "production"
+        uuid: "UUID1",
```
EdwinGuzman: None of this test data is used in this file, no?

src/components/reusables/Tabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+export interface TabsState {
+  tab: string;
+}
+
+export default class Tabs extends React.Component<TabsProps, TabsState> {
+  constructor(props: TabsProps) {
+    super(props);
+    this.select = this.select.bind(this);
+    this.makeTabs = this.makeTabs.bind(this);
+    this.state = { tab: Object.keys(this.props.items)[0] };
```
EdwinGuzman: Saving the current state as a string and the title of the panel does not feel very stable.

src/components/reusables/Tabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+      );
+      navs.push(navItem);
+      let contentItem = <section key={item[0]} className={`tab-content ${hidden}`}>{item[1]}</section>;
+      content.push(contentItem);
+    });
+    return [navs, content];
+  }
+
+  render(): JSX.Element {
+    let navs = this.makeTabs()[0];
```
EdwinGuzman: You should deconstruct this because you are currently calling `this.makeTabs()` twice for each library. That means you are constructing the nav and content twice but only extracting what you need from each one.

src/components/reusables/Tabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  makeTabs(): Array<Array<JSX.Element>> {
+    let navs = [] as Array<JSX.Element>;
+    let content = [] as Array<JSX.Element>;
+    Object.entries(this.props.items).map(item => {
+      let hidden = item[0] === this.state.tab ? "" : "hidden";
+      let current = item[0] === this.state.tab ? "current" : "";
+      let navItem = (
+        <li key={item[0]} className={`tab-nav ${current}`}>
```
EdwinGuzman: This li needs a role of "presentation".

src/components/reusables/Tabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <button className="btn btn-default" onClick={this.select}>
```
EdwinGuzman: This button needs a role of "tab" as well as keeping track if it is selected or not with `aria-selected`.

src/components/reusables/Tabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+            {item[0]}
+          </button>
+        </li>
+      );
+      navs.push(navItem);
+      let contentItem = <section key={item[0]} className={`tab-content ${hidden}`}>{item[1]}</section>;
```
EdwinGuzman: This section is a panel and should have a role of "tabpanel".
This also needs an `aria-labelledby` pointing to the id of the associated tab `button`.

src/components/reusables/Tabs.tsx
EdwinGuzman: This also needs an id for another issue down below.

src/components/reusables/Tabs.tsx
EdwinGuzman: This also needs a tabindex of 0 so you can tab into it. What you want to do is tab into the panel from the tab instead of tabbing between tabs. You want to use the left and right arrows to navigate between tabs.

src/components/LibraryDetailPage.tsx
EdwinGuzman: Never mind, it wasn't updated for me at first but now I see the update 👍

src/components/reusables/Tabs.tsx
adriana-alter: What do you think would be better?

src/components/__tests__/LibrariesList-test.tsx
adriana-alter: No.  But TypeScript becomes distressed if I don't have it (because the interface is expecting all of this).

src/components/LibraryDetailPage.tsx
adriana-alter: What do you mean?

src/components/LibraryDetailPage.tsx
adriana-alter: How so?

src/components/LibraryDetailPage.tsx
EdwinGuzman: The interface wasn't updated when I checked at first but later on it was fine. Weird.

src/components/LibraryDetailPage.tsx
EdwinGuzman: If you were to test renderItems in isolation, you cannot because it depends on `this.props` for the library data.

src/components/reusables/Tabs.tsx
EdwinGuzman: And id of some kind.

src/components/__tests__/LibrariesListItem-test.tsx
```diff
@@ -58,11 +67,11 @@ describe("LibrariesListItem", () => {
       expect(wrapper.state().color).to.equal("warning");
       expect(wrapper.find(".panel-warning").length).to.equal(1);
 
-      (wrapper.instance() as any).updateColor("cancelled", "production");
+      (wrapper.instance() as any).updateColor(["cancelled", "production"]);
```
EdwinGuzman: Which one is registry stage and library stage? It'll be clearer to have them as variable names so you can see which param is which stage.

src/components/__tests__/LibraryDetailPage-test.tsx
```diff
@@ -44,19 +52,26 @@ describe("LibraryDetailPage", () => {
...
+    expect(contactUrlItems.length).to.equal(4);
+
+    let allItems = Object.assign(basicInfoItems, contactUrlItems);
+    let libraryData = Object.assign(library.basic_info, library.urls_and_contact);
+    allItems.map((item: Enzyme.CommonWrapper<any, any, {}>) => {
       let key = item.find(".control-label").text();
-      let value = item.find(".form-control-static").text();
-      expect(`${library[key]}`).to.equal(value);
+      let value = item.find("span").text();
+      expect(`${libraryData[key]}`).to.equal(value);
```
EdwinGuzman: The key value is already a string so you don't need to extrapolate this variable.

src/components/__tests__/LibrariesList-test.tsx
EdwinGuzman: You can remove the ones that are optional.

adriana-alter:
I made it so you can tab into the panel from the tab...but (many hours and many online tutorials later!) I still haven't gotten the arrow navigation to work.

src/components/__tests__/LibrariesListItem-test.tsx
adriana-alter: Does it matter which is which, though?  The color-coding logic is just based on the combination of the two stages--it checks 1) whether either stage is cancelled, and 2) whether both stages are in production.  The individual values should only matter for `LibraryDetailPage` and its children (i.e. for populating and submitting the form); if `LibrariesListItem` suddenly cares about which is which, then something has gone wrong...

src/stylesheets/colors.scss
```diff
@@ -8,9 +8,13 @@ $red-dark: #97272C;
 $light-gray: #D7D4D0;
 $dark-gray: #080807;
 $gray-tint: tint($light-gray, 75);
+$gray: #76777B;
+$gray-brown: #54514A;
 $black: #111;
 $blue: #1B7FA7;
-$light-blue: lighten($blue, 30%);
+$blue-dark: #135772;
+$light-blue: #5FA5C1;
```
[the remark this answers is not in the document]
adriana-alter: They're definitely supposed to be from the design toolkit--I may have accidentally flipped a few names, though.  Will check and fix!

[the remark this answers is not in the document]
adriana-alter: I made `gray-tint`, based off of `light-gray` from the toolkit.

I just noticed that the toolkit website has a typo--it lists the same hex codes for `gray` and `gray-brown`.  The right code for `gray` is "#776e64".  Is there someone we should point this out to?

Yes, I flipped/mistyped the blue ones!

Most of the other things that are lightened or darkened are from the toolkit, but I'll write out the hex codes to make that clearer.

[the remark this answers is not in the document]
EdwinGuzman: Cool. You can create a Github issue in the Design Toolkit repo.

src/components/reusables/Tabs.tsx
```diff
@@ -26,19 +26,43 @@ export default class Tabs extends React.Component<TabsProps, TabsState> {
...
     let content = [] as Array<JSX.Element>;
-    Object.entries(this.props.items).map(item => {
+    let items = Object.entries(this.props.items);
+
+    items.map(item => {
       let [name, data] = item;
-      let hidden = name === this.state.tab ? "" : "hidden";
-      let current = name === this.state.tab ? "current" : "";
+      let current = name === this.state.tab;
+      let idx = items.indexOf(item);
```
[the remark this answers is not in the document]
adriana-alter: Yes, good point!

adriana-alter:
Okay, so:

I still can't get the arrow navigation to work.  (It doesn't work in similar components in circ-web either, for what that's worth.) . Unless there's something very obvious that I'm missing because I've been staring at it for too long, it does not seem worth holding up this branch just because of that; we can always do an accessibility improvements branch later on.  (And tabbing into the panel does work.)

EdwinGuzman:
Ooh I'd say you'd have to keep state for which tab is selected and you can read the key press value and check if it is left or right. Then you'd move through the tab and focus on it.

Accessibility is important and since this PR is for the Tabs feature as a whole, it should be included. For next time we can break it out and you can work on different aspects of a feature. In general, I would not feel comfortable merging in a feature to `master` that may work but is not complete, and missing this accessibility feature makes it not complete.

EdwinGuzman:
For next time, the git workflow can be something like:
```
 ---master-- ---- ----- --------- -- ------
        \                                                     /
         Tabs ---- ---- ----- --------- -- 
                       \      \            /              /
                       UI ---\ -- ---              /
                                Accessibility---/  
```

I hope that makes sense. Basically, you'd create a feature branch `Tabs` where you would merge (as PRs) different parts of the WIP until it's ready to be a PR into `master`. We've done it for other projects here and makes it better to merge in a feature once it's all ready for `master`.

adriana-alter:
Got the arrow navigation working!

I made it so that, if you're already on the last tab and you hit the right arrow, you loop back to the first tab (and if you're already on the first tab and you hit the left arrow, you loop back to the last tab).  Does that seem like reasonable behavior, or should you just stay on the current tab if there's nowhere to go?

src/components/__tests__/LibrariesList-test.tsx
adriana-alter: These are already the ones that aren't optional...

EdwinGuzman:
Sweet. This is much better and conforming to the WCAG standard  👌 
I think you just need updated tests right?

adriana-alter:
Cool!  Do you think the comments in `select` make sense, or would it be more readable if I had separate `onClick` and `onKeyDown` functions?

I need 1) updated tests for Tabs, and 2) to go back and look at what you were saying about making `renderItems` more testable.  Stay tuned!

EdwinGuzman:
It makes sense to me but I can see how it would be more readable if they were separated into two functions for the `onClick` and the `onKeyDown` functions. That's up to you.

EdwinGuzman:
I wouldn't worry too much about 2 if you can't get it. It's either testing the function through the component or testing the function individually. Individually would be nice so you can also notice how coupled (or not) a function is to the component or if the function is pure and can rely just on its params.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ###v0.0.3
 #### Updated
-- Organized library details into tabs
+- Made library details more readable by implementing an accessible Tabs component
```
EdwinGuzman: You haven't published `0.0.2` yet so technically you can combined this PR with `0.0.2` if you want to.

src/components/LibraryDetailPage.tsx
```diff
@@ -88,15 +86,24 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
...
     }
+    let library = this.props.fullLibrary || this.props.library;
+    let tabItems = {};
+
+    const categories = {
+      "Basic Information": "basic_info",
+      "Contact & URLs": "urls_and_contact"
+    };
+
+    Object.keys(categories).map(tabName => {
```
EdwinGuzman: `each` would be better than `map` since you don't need to create a new array.

PR #4, opened 2019-01-28
Login

adriana-alter:
-- Shows log in form if user isn't logged in.  (NOTE: credentials are "Admin", "123")
-- Shows log-out link in header
-- Implements reusable form, fieldset, and input components


EdwinGuzman:
Fixes [SIMPLY-1574](https://jira.nypl.org/browse/SIMPLY-1574).

EdwinGuzman:
I keep getting this error when building and running the server locally, even on a clean `npm install` step:
screen shot 2019-01-29 at 12 10 22 pm

EdwinGuzman:
* The log in screen is not centered
screen shot 2019-01-29 at 12 16 36 pm
* There are no error messages when I mistype the username or password
* The log out button should have "log out" next to it (for accessibility - in general you should not let only a color or icon dictate meaning so it should be accompanied with text)
screen shot 2019-01-29 at 12 16 44 pm

adriana-alter:
This branch fixed that error for me.  Hmm.  I will investigate.

The log in screen was not intended to be centered!  (I can center it if you think that would look better, though.)

Error handling--for this and other things--is going to be in future PRs.

Will add log out text!

EdwinGuzman:
Hmm I think the error message is important to the log in form (or any form in general). If it will be a later PR then this PR cannot be `1.0.0` since then it will still be missing key features.

EdwinGuzman:
The log in form looks better centered like the Circ Admin but if you don't want to, that's fine.

adriana-alter:
Ok--I'll do the error message on a new branch, and that can be 1.0.0.  This PR is already way too big.

src/__tests__/index-test.ts
```diff
@@ -1,34 +1,46 @@
...
 import * as React from "react";
 import * as ReactDOM from "react-dom";
-import { mount } from "enzyme";
+import * as Enzyme from "enzyme";
+import * as Sinon from "sinon";
 
 const RegistryAdmin = require("../index");
 import ContextProvider from "../components/ContextProvider";
-import { Router } from "react-router";
+import { Router, Route } from "react-router";
```
EdwinGuzman: I don't think you're using `Route`.

adriana-alter:
Do you think the credentials fieldset and the button should be centered inside the login form?

src/__tests__/index-test.ts
```diff
@@ -1,34 +1,46 @@
...
     const provider = wrapper.find(ContextProvider);
     expect(provider.length).to.equal(1);
     renderSpy.restore();
   });
 
   it("renders Router", () => {
-    const renderSpy = spy(ReactDOM, "render");
-    new RegistryAdmin({});
+    const renderSpy = Sinon.spy(ReactDOM, "render");
+    new RegistryAdmin({ username: "Admin" });
```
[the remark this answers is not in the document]
EdwinGuzman: And you only need a username and not username and password to be logged in?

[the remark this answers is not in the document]
adriana-alter: If the password is missing/wrong, the login attempt has already failed on the server side...

src/components/reusables/Fieldset.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react";
+
+export interface FieldsetProps {
+  legend: string;
+  elements: Array<JSX.Element>;
+  labelClass?: string;
+  labelText?: string;
```
EdwinGuzman: I was going to say that `<label>`s should always be required when using any form elements, but you're using it differently. You don't need the `<label>` element inside the `<legend>` since the legend is already a label for the fieldset. This is confusing because you have two props that are both labels inside the legend, but I think the legend should have a more static label to describe the form instead of displaying information - if that makes sense. So something like "Library Stage" and the "current library stage" can be an element outside the label so it doesn't describe the following form. Missed this the first time.

src/__tests__/index-test.ts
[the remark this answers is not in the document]
EdwinGuzman: This just seems weird to rely on the username to load the app.

src/components/reusables/Form.tsx
```diff
@@ -23,8 +27,12 @@ export default class Form extends React.Component<FormProps, void> {
   };
 
   render(): JSX.Element {
+    let className = this.props.className ? `clearfix ${this.props.className}` : "clearfix";
```
EdwinGuzman: If `clearfix` will always be rendered, you could have it in the `className` attribute below.

src/components/reusables/Header.tsx
```diff
@@ -1,13 +1,16 @@
...
 export interface HeaderProps {
   text?: string;
   imgSrc?: string;
   alt?: string;
+  logOut?: string;
 }
 
 export default class Header extends React.Component<HeaderProps, void> {
   render(): JSX.Element {
+    let src = this.props.imgSrc ? require(`${this.props.imgSrc}`) : "";
```
EdwinGuzman: You are not actually using this variable. I think you want to use it on line 20.

src/__tests__/index-test.ts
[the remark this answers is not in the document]
adriana-alter: I think it's okay for it to be weird, given that the server-side login implementation is weird anyway (only one account, and hard-coded credentials)...

src/components/reusables/LogInForm.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  render(): JSX.Element {
+    let title = this.props.title ? this.props.title : "Log In";
+    let legend = this.props.legend ? this.props.legend : "Credentials";
+
+    let username = <Input key="username" name="username" label="Username" />;
+    let password = <Input key="password" type="password" name="password" label="Password" />;
+    let elements = this.props.extraFields ? [username, password].concat(this.props.extraFields) : [username, password];
+    let fieldset = <Fieldset key="credentials" legend={legend} elements={elements} />;
```
EdwinGuzman: Why do you need these `key` props here?

src/components/reusables/LogInForm.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Store } from "redux";
+import { AdminData } from "../../interfaces";
+import Form from "./Form";
+import Fieldset from "./Fieldset";
+import Input from "./Input";
+import ActionCreator from "../../actions";
+import { State } from "../../reducers/index";
+
+export interface LogInFormStateProps {
+  admin?: AdminData;
```
EdwinGuzman: Don't think you are using this.

src/components/reusables/__tests__/Fieldset-test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    expect(legend.length).to.equal(1);
+    expect(legend.text()).to.equal("TEST_LEGEND");
+  })
+  it("displays elements", () => {
+    let inputs = wrapper.find("input");
+    expect(inputs.length).to.equal(2);
+    expect(inputs.at(0).prop("value")).to.equal("A");
+    expect(inputs.at(1).prop("value")).to.equal("B");
+  });
+  it("optionally displays a label", () => {
```
EdwinGuzman: This is confusing because you can "optionally" display a label but you are always displaying a legend. The label is just extra text at the end of the legend. I don't think the names match up semantically.

src/components/reusables/__tests__/Fieldset-test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    wrapper.setProps({ labelText: "here's a label!" });
+    let label = wrapper.find("legend").find("label");
+    expect(label.length).to.equal(1);
+    expect(label.text()).to.equal("here's a label!");
+    expect(label.hasClass("label-default")).to.be.true;
+  });
+  it("optionally specifies a class for the label", () => {
+    wrapper.setProps({ labelText: "Success!", labelClass: "success" });
```
EdwinGuzman: Like I mentioned above, the legend for the form you are using is dynamic and I'm not sure if that's the right way to use a legend for a fieldset.

src/components/reusables/LogInForm.tsx
adriana-alter: Because React becomes distressed if child components don't have unique `key` props.

src/components/reusables/__tests__/Input-test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    let input = wrapper.find("input");
+    expect(input.prop("name")).to.equal("inputName");
+    expect(input.prop("id")).to.equal("inputName");
+  });
+  it("defaults to setting the input's type as text", () => {
+    let input = wrapper.find("input");
+    expect(input.prop("type")).to.equal("text");
+  });
+  it("optionally accepts a type prop", () => {
+    wrapper.setProps({ type: "radio" });
```
EdwinGuzman: This is fine, but I would explicitly create a new wrapper like `wrapper = Enzyme.shallow(<Input type="radio" ...`

src/components/reusables/LogInForm.tsx
EdwinGuzman: Hmm I understand for the username and password for the array but not for the fieldset. I don't get any errors when removing the `key` prop.

src/index.tsx
```diff
@@ -3,12 +3,14 @@ import * as ReactDOM from "react-dom";
...
 import ContextProvider from "./components/ContextProvider";
 import App from "./components/App";
+import LogInForm from "./components/reusables/LogInForm";
 
 interface ConfigurationSettings {
   /** A token generated by the server to prevent Cross-Site Request Forgery.
       The token should be included in an 'X-CSRF-Token' header in any non-GET
       requests. */
-  csrfToken: string;
+  csrfToken?: string;
```
EdwinGuzman: You're still not using the `csrfToken`?

EdwinGuzman: It's still commented out in the `ContextProvider` and I thought it was going to be used for logging in?

src/components/reusables/LogInForm.tsx
adriana-alter: I get errors when I remove it.  (Try logging out.)

src/actions.ts
```diff
@@ -17,13 +17,9 @@ export default class ActionCreator extends BaseActionCreator {
 
   static readonly LOG_IN = "LOG_IN";
 
-  csrfToken: string;
-
-  constructor(fetcher?: DataFetcher, csrfToken?: string) {
+  constructor(fetcher?: DataFetcher) {
```
EdwinGuzman: Are you sure you don't need the `csrfToken`? This is something I'm not too familiar with and I'd ask Amy for help regarding making safe calls to the server for this.

src/components/reusables/Form.tsx
```diff
@@ -33,8 +40,10 @@ export default class Form extends React.Component<FormProps, void> {
           />
         }
         { this.props.content }
-        <SaveButton
+        <br />
```
EdwinGuzman: No way to not use a `<br />`? I find them harder to maintain than updating style rules in css.

src/components/reusables/LogInForm.tsx
EdwinGuzman: Hmm yeah I see them.

EdwinGuzman:
Very nice. The update looks good but I'm just unsure about the `csrfToken` and the text + label in the legend for a fieldset. It doesn't seem to be semantically correct since it's dynamic but probably doesn't matter 👍 .

I also get this error in an incognito window (did try to go to `/admin/log_out`):
screen shot 2019-01-29 at 3 37 18 pm

adriana-alter:
Working on the legend/label thing right now :)

Well, I was sure the CSRF token thing was fine, but that error message is giving me pause...

src/components/reusables/Form.tsx
adriana-alter: I think the `<br />` is more likely to work no matter what kind of elements someone has passed to the Form component.

adriana-alter:
It looks to me like it would be okay to just remove the CSRF stuff from the server code (which would get rid of that error message, and--to my knowledge--not harm anything), but that seems potentially drastic.  @aslagle, what do you think?

aslagle:
There isn’t a lot of risk right now since only a couple people will use it and the only thing they can do is change stages, so I think it’s fine to remove it and implement it later if more dangerous functionality is added

adriana-alter:
Great--thanks, Amy!

EdwinGuzman:
Last update looks good. I'll just retest when the server part is updated.

adriana-alter:
@EdwinGuzman Server part is updated!

EdwinGuzman:
I constantly get this error message when I open a new incognito window now:
screen shot 2019-01-30 at 3 53 32 pm

If I go to `/admin/log_out`, then I can see the log in form. But going to that url is not a good solution..

adriana-alter:
That is...not good.  I will investigate.

PR #6, opened 2019-01-30
Login errors

adriana-alter:
screen shot 2019-01-30 at 12 41 29 pm

### DIFF
--- a/src/stylesheets/reusables/form.scss
+++ b/src/stylesheets/reusables/form.scss
@@ -22,6 +22,12 @@ form {
     }
   }
 
+  .alert.alert-danger {
+    background-color: $red-error;
+    color: $dark-gray;
+    font-weight: bold;
+  }
+
   .form-title {
     display: block;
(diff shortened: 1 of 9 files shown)


src/stylesheets/reusables/form.scss
```diff
@@ -22,6 +22,12 @@ form {
     }
   }
 
+  .alert.alert-danger {
```
EdwinGuzman: Can you make this a child instead?

EdwinGuzman:
Looks good. Just make it focus on the error message when it appears and you're all set.

src/stylesheets/reusables/form.scss
adriana-alter: What do you mean--make it a child of what?

EdwinGuzman:
````
  .alert {
    &.alert-danger { 
     .....
```

PR #7, opened 2019-02-12
Add feature for re-sending validation email

adriana-alter:
The relevant server code for this has already been merged.

screen shot 2019-02-12 at 1 24 09 pm
screen shot 2019-02-12 at 1 23 54 pm
screen shot 2019-02-12 at 1 23 32 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 ## Changelog
 
-####v1.0.0
+###v1.0.1
+#### Added
+- Implemented form for re-sending validation emails 
+
+###v1.0.0
 #### Added
 - Log in functionality
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,20 +1,20 @@
 ## Changelog
 
-###v1.0.1
+### v1.0.1
 #### Added
-- Implemented form for re-sending validation emails 
+- Implemented form for re-sending validation emails
 
-###v1.0.0
+### v1.0.0
 #### Added
 - Log in functionality
 
-###v0.0.2
+### v0.0.2
 #### Added
 - Rendered list of libraries
 - Rendered details about each library
 - Implemented functionality for editing library stage and registry stage
 - Made library details more readable by implementing an accessible Tabs component
 
-###v0.0.1
+### v0.0.1
 #### Added
 - Set up the interface
--- a/src/components/EmailForm.tsx
+++ b/src/components/EmailForm.tsx
@@ -37,18 +37,17 @@ export class EmailForm extends React.Component<EmailFormProps, EmailFormState> {
 
   async sendEmail(data: FormData): Promise<void> {
     await this.props.email(data);
-    this.setState({ sent: true });
+    this.setState({ sent: !this.props.error });
   }
 
   render(): JSX.Element {
-    let disabled = false;
-    let icon = this.state.sent ? <CheckSoloIcon iconId="1" /> : null;
+    let icon = this.state.sent ? <CheckSoloIcon /> : null;
     let hasEmail = !!this.props.library.urls_and_contact.contact_email;
     let alreadyValidated = this.props.library.urls_and_contact.validated !== "Not validated";
 
     let buttonText = hasEmail ? "Send Validation Email" : "No email address configured";
     let buttonContent = <span>{buttonText}{icon}</span>;
-    let infoText = alreadyValidated && !this.state.sent && !this.props.error && "Already validated";
+    let infoText = (alreadyValidated && !this.state.sent && !this.props.error) ? "Already validated" : null;
     let successText = `Validation email successfully sent to ${this.props.library.urls_and_contact.contact_email}`;
 
     return (
@@ -60,7 +59,7 @@ export class EmailForm extends React.Component<EmailFormProps, EmailFormState> {
           onSubmit={this.sendEmail}
           buttonContent={buttonContent}
           disableButton={!hasEmail}
-          infoText={alreadyValidated ? infoText : null}
+          infoText={infoText}
           errorText={this.props.error ? this.props.error.response : null}
           successText={this.state.sent ? successText : null}
         />
--- a/src/components/EmailForm.tsx
+++ b/src/components/EmailForm.tsx
@@ -1,0 +1,89 @@
+import * as React from "react";
+import Form from "./reusables/Form";
+import { connect } from "react-redux";
+import ActionCreator from "../actions";
+import { FetchErrorData } from "opds-web-client/lib/interfaces";
+import { Store } from "redux";
+import { State } from "../reducers/index";
+import { LibraryData } from "../interfaces";
+import { CheckSoloIcon } from "@nypl/dgx-svg-icons";
+
+export interface EmailFormState {
+  sent: boolean;
+}
+
+export interface EmailFormOwnProps {
+  library: LibraryData;
+  store: Store<State>;
+}
+
+export interface EmailFormDispatchProps {
+  email: (data: FormData) => Promise<void>;
+}
+
+export interface EmailFormStateProps {
+  error?: FetchErrorData;
+}
+
+export interface EmailFormProps extends EmailFormOwnProps, EmailFormDispatchProps, EmailFormStateProps {}
+
+export class EmailForm extends React.Component<EmailFormProps, EmailFormState> {
+
+  constructor(props) {
+    super(props);
+    this.sendEmail = this.sendEmail.bind(this);
+    this.state = { sent: false };
+  }
+
+  async sendEmail(data: FormData): Promise<void> {
+    await this.props.email(data);
+    this.setState({ sent: true });
+  }
+
+  render(): JSX.Element {
+    let disabled = false;
+    let icon = this.state.sent ? <CheckSoloIcon iconId="1" /> : null;
+    let hasEmail = !!this.props.library.urls_and_contact.contact_email;
+    let alreadyValidated = this.props.library.urls_and_contact.validated !== "Not validated";
+
+    let buttonText = hasEmail ? "Send Validation Email" : "No email address configured";
+    let buttonContent = <span>{buttonText}{icon}</span>;
+    let infoText = alreadyValidated && !this.state.sent && !this.props.error && "Already validated";
+    let successText = `Validation email successfully sent to ${this.props.library.urls_and_contact.contact_email}`;
+
+    return (
+        <Form
+          className="validation"
+          hiddenName="uuid"
+          hiddenValue={this.props.library.uuid}
+          title="Validation"
+          onSubmit={this.sendEmail}
+          buttonContent={buttonContent}
+          disableButton={!hasEmail}
+          infoText={alreadyValidated ? infoText : null}
+          errorText={this.props.error ? this.props.error.response : null}
+          successText={this.state.sent ? successText : null}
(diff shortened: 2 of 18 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 ## Changelog
 
-####v1.0.0
+###v1.0.1
```
EdwinGuzman: You need spaces between the hashtag and the v or else the markdown doesn't format correctly.

EdwinGuzman:
In the example above, it's a bit weird that you have `a` and `123` as contact emails since the Circ Admin doesn't validate and allow those values as emails. The stage and basic information areas are both aligned left so it's a bit weird to see this validation section aligned center.

src/components/EmailForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { LibraryData } from "../interfaces";
+import { CheckSoloIcon } from "@nypl/dgx-svg-icons";
+
+export interface EmailFormState {
+  sent: boolean;
+}
+
+export interface EmailFormOwnProps {
+  library: LibraryData;
+  store: Store<State>;
```
EdwinGuzman: You don't seem to be using the store in this component.

src/components/EmailForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  async sendEmail(data: FormData): Promise<void> {
+    await this.props.email(data);
+    this.setState({ sent: true });
+  }
+
+  render(): JSX.Element {
+    let disabled = false;
+    let icon = this.state.sent ? <CheckSoloIcon iconId="1" /> : null;
```
EdwinGuzman: The `CheckSoloIcon` doesn't use `iconId` but it's good you tried to add it. If it was working, you shouldn't use numbers for ids like this. Also, screenreaders will give you a "duplicate id on page" error if they read multiple elements with the same id. In this case, I think it'll happen since you would have X number of validation forms for X number of libraries.

src/components/EmailForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+    this.state = { sent: false };
+  }
+
+  async sendEmail(data: FormData): Promise<void> {
+    await this.props.email(data);
+    this.setState({ sent: true });
+  }
+
+  render(): JSX.Element {
+    let disabled = false;
```
EdwinGuzman: You're not using this variable.

src/components/EmailForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+    return (
+        <Form
+          className="validation"
+          hiddenName="uuid"
+          hiddenValue={this.props.library.uuid}
+          title="Validation"
+          onSubmit={this.sendEmail}
+          buttonContent={buttonContent}
+          disableButton={!hasEmail}
+          infoText={alreadyValidated ? infoText : null}
```
EdwinGuzman: If `alreadyValidated` is false, then `infoText` above will also be undefined/false, no? It seems you're doing similar checks to check for validation and then rendering the info text.

src/components/LibraryDetailPage.tsx
```diff
@@ -40,15 +41,15 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
...
+      registryStage: this.props.library.stages.registry_stage,
     };
   }
 
   renderItems(category: {[key: string]: string}): JSX.Element {
 
     // Only create LibraryDetailItems for fields which actually have a value.
     let fields = Object.keys(category).filter(label => category[label]).map(label =>
-      <LibraryDetailItem key={label} label={label} value={category[label]} />
+      <LibraryDetailItem key={label} label={label} value={`${category[label]}`} />
```
EdwinGuzman: Why this update?

src/components/LibraryDetailPage.tsx
```diff
@@ -79,7 +80,7 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
     let libraryStage = this.props.fullLibrary.stages.library_stage;
     let registryStage = this.props.fullLibrary.stages.registry_stage;
     this.props.updateColor([libraryStage, registryStage]);
-    this.setState({ libraryStage, registryStage });
+    this.setState(Object.assign(this.state, { libraryStage, registryStage }));
```
EdwinGuzman: Why this update?

src/components/reusables/Form.tsx
```diff
@@ -35,21 +39,25 @@ export default class Form extends React.Component<FormProps, void> {
     this.props.onSubmit(data);
   };
 
-  errorMessage(errorText: string): JSX.Element {
-    let error = (
-      <p className="alert alert-danger" role="alert" ref="errorMessage" tabIndex={-1}>
-        Error: {errorText}
+  message(text: string, type: string): JSX.Element {
```
EdwinGuzman: Very nice refactoring of this to accommodate types of messages.

src/components/EmailForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+export class EmailForm extends React.Component<EmailFormProps, EmailFormState> {
+
+  constructor(props) {
+    super(props);
+    this.sendEmail = this.sendEmail.bind(this);
+    this.state = { sent: false };
+  }
+
+  async sendEmail(data: FormData): Promise<void> {
+    await this.props.email(data);
```
EdwinGuzman: This action only calls the server to send an email, but that action itself could return an error. You're catching that error in a state prop, but it seems like the state for `sent` will still be true. I don't see any place where it is being set to false so is it possible that `sent` will be true even though you get an error message from the server? I can't readily test this scenario and hope you have.

src/components/__tests__/EmailForm-test.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+    const pause = (): Promise<void> => {
+      return new Promise<void>(resolve => setTimeout(resolve, 0));
+    };
+    await pause();
+
+    expect(wrapper.state()["sent"]).to.be.true;
+  });
+
+  it("displays a success message", () => {
```
EdwinGuzman: If you set `sent`'s state to true and also pass in the error prop, both the success and danger messages will show up at the same time. So might be good to check for that case in the component.

src/components/reusables/__tests__/Form-test.tsx
```diff
@@ -93,6 +130,14 @@ describe("Form", () => {
     expect(button.props()["submit"]).to.equal((wrapper.instance() as any)["save"]);
   });
 
+  it("should optionally props to the SubmitButton", () => {
```
EdwinGuzman: Optionally "send" props?

src/components/EmailForm.tsx
adriana-alter: It breaks if I take the store out...

src/components/reusables/__tests__/Form-test.tsx
adriana-alter: Good catch :)

adriana-alter:
Here's how the validation section looks if I align it left--that looks weirder to me than aligning it center does.  What do you think?
screen shot 2019-02-13 at 9 55 32 am

I agree that it looks weird that the sample data in my local database happens to have non-valid email addresses, but not worth messing with that just for the sake of these screenshots.  (The circ manager doesn't have a problem with non-valid email addresses that were created before we implemented the address validator; that's where these are coming from.  It's not a sign that anything is wrong.)  And if we ever need screenshots for documentation or something, we'd do them from QA or production anyway, not from local.

src/components/LibraryDetailPage.tsx
adriana-alter: Because of something that I subsequently got rid of.  I'll revert this!

src/components/EmailForm.tsx
adriana-alter: My understanding is that, if the action results in an error, the subsequent code (i.e. the `setState` part) won't be run.  `LogInForm` does something very similar (line 43 et seq.)--if the `logIn` action fails, it will just show the error message, not go on to reload the page.  Wrong?

But at any rate, even if that's not the right explanation: I did test it, and it works the way it's supposed to; it does not set `sent` to true if there's an error message.

src/components/__tests__/EmailForm-test.tsx
adriana-alter: I will go ahead and put in a safeguard just in case, but I don't see any way for this situation to arise in real life.  (See above: if there's an error prop, `sent` never gets set to true.)

adriana-alter: Can you think of any situation (e.g. if there were multiple fieldsets in the same form) in which, realistically, we would want a form to simultaneously display an error message AND a success message?  (I'm considering putting the safeguard into the `Form` component--making it so that it can only show one at a time.)

src/components/EmailForm.tsx
EdwinGuzman: I took it out and it didn't break for me.

src/components/LibraryDetailPage.tsx
EdwinGuzman: Hmm sounds good. Except values being boolean or numbers doesn't seem right...

src/components/LibraryDetailPage.tsx
EdwinGuzman: Alright. It also doesn't do anything since `this.state` only has two values and you are overriding those values anyway.

adriana-alter: I know--the thing I got rid of involved having a third value :)

src/components/EmailForm.tsx
EdwinGuzman: Hm I figured it'd both go to the next step if it's successful or fails. But if it works...

src/components/__tests__/EmailForm-test.tsx
EdwinGuzman: I definitely do not see any case were we would want a form to display both, but it seemed like the logic would've caused both to show up. But if the `setState` function does not get called after the async function, then it's fine.

EdwinGuzman:
I think it looks better. Why not ask Jack for input?
I figured it was previously added data that wasn't validated before.

src/components/LibraryDetailPage.tsx
adriana-alter: I think it's fine for `online_registration` to be a boolean (especially since it's the only one).  I agree that numbers would be slightly weird.  But we can reevaluate that if/when it comes up in the future.

adriana-alter:
I messed with it very slightly--I think this looks nicer.
screen shot 2019-02-13 at 2 23 23 pm

EdwinGuzman:
I agree it looks nicer.

adriana-alter:
When I remove `store` from `EmailForm`, the app runs fine, but the tests crash: when it tries to run the `beforeEach` hooks for each of `LibraryDetailPage-test`, `LibrariesListItem`, and `LibraryList`, it generates an error like so: ` Invariant Violation: Could not find "store" in either the context or props of "Connect(EmailForm)". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "Connect(EmailForm)".`

I don't know where this is coming from; as far as I can tell, I've removed all of the references to `EmailForm` needing a store...

EdwinGuzman:
Hmm I guess because of the `connect`. 👍

PR #8, opened 2019-02-15
Update names and copy to reflect new email functionality

adriana-alter:
I changed the email validation feature such that, instead of sending a validation email, it manually validates the email address.  The intended use case involves a situation in which we already know that the email address is valid, but the validation link has expired because the recipient didn't click on it in time; it makes more sense to manually validate the email address than to modify the server-side logic for determining which type of email to send, as it turns out we would need to do.  The actual functionality for this is in the server code (already merged); this branch is just updating some copy and variable names to reflect the changes.

https://jira.nypl.org/browse/SIMPLY-1573

### DIFF
--- a/src/__tests__/actions-test.ts
+++ b/src/__tests__/actions-test.ts
@@ -327,8 +327,8 @@ describe("actions", () => {
     });
   });
 
-  describe("email", () => {
-    it("sends an email", async () => {
+  describe("validate_email", () => {
+    it("validates an email address", async () => {
       const dispatch = stub();
       const formData = new (window as any).FormData();
       formData.append("uuid", "abc");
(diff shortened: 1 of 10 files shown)


EdwinGuzman:
Why the switch from sending an email validation to manually validating the email address?

src/__tests__/actions-test.ts
```diff
@@ -338,11 +338,11 @@ describe("actions", () => {
...
       fetch = fetchMock;
 
-      await actions.email(formData)(dispatch);
+      await actions.validate_email(formData)(dispatch);
       expect(dispatch.callCount).to.equal(2);
 
-      expect(dispatch.args[0][0].type).to.equal("EMAIL_REQUEST");
-      expect(dispatch.args[1][0].type).to.equal("EMAIL_SUCCESS");
+      expect(dispatch.args[0][0].type).to.equal("VALIDATE_EMAIL_REQUEST");
+      expect(dispatch.args[1][0].type).to.equal("VALIDATE_EMAIL_SUCCESS");
```
EdwinGuzman: Hmm there's no way the server can fail and you could get an error message back instead of a success?

adriana-alter: There is not.  The only way that something could hypothetically go wrong with validating an email is that there might not be an email address associated with the library...but in that case, the button would be disabled, so it would be impossible to get to this point in the first place.

PR #9, opened 2019-02-19
Update library info after validating email, and change variable name

adriana-alter:
https://jira.nypl.org/browse/SIMPLY-1725

Use with https://github.com/NYPL-Simplified/library_registry/pull/102 to get rid of the weird flash when the info updates.

Added a call to `fetchLibrary` after email address is validated.

(Also, I changed the variable name `validate_email` to `validateEmail`, just because JS syntax.)

Update: discovered a bug whereby `fetchLibrary` was causing all of the `LibraryDetailPage` tab components to display the info for the library that gets fetched.  Fixed it by checking the uuid in `LibraryDetailPage.tsx`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v1.0.3
+#### Updated
+- Made the validation timestamp automatically update as soon as a user validates an email address 
+
 ### v1.0.2
 #### Updated
 - Implemented functionality for manually validating an email address
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-registry-admin",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "description": "Web Front-end for Library Simplified Circulation Manager Admin Interface",
   "repository": {
     "type": "git",
--- a/src/__tests__/actions-test.ts
+++ b/src/__tests__/actions-test.ts
@@ -327,7 +327,7 @@ describe("actions", () => {
     });
   });
 
-  describe("validate_email", () => {
+  describe("validateEmail", () => {
     it("validates an email address", async () => {
       const dispatch = stub();
       const formData = new (window as any).FormData();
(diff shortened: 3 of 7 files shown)


EdwinGuzman:
The code itself looks fine but there's a bug which I don't know if it happens on the client side or the server side (which I haven't reviewed most of).

Yes, your last commits makes sure that only the relevant library's info gets updated in _real time_, but if you refresh the page, all the libraries' `validated` field get updated to the same timestamp as the library's email you just validated. So it seems that somehow all the emails from each library are validated. Your change makes sure that the relevant library you're looking at is updated, but actually all of the libraries are updated.

EdwinGuzman:
I want to point out that this issue is happening on the `master` branch as well, so I don't think the bug is coming from this PR _but_ since you are updating that specific feature, it's worth also noting this information.

adriana-alter:
Agreed, the bug is not coming from this PR; I just happened to have occasion to notice it while I was working on this PR.

adriana-alter:
The only situation in which I can reproduce the bug is if multiple libraries have the same email address.  Are you able to reproduce it a different way...?

EdwinGuzman:
Yeah, it's not from this PR but I was trying to say that _that_ should be fixed instead (is a more important bug) than refreshing the data right away.

EdwinGuzman:
I'll check (re: same email), but you're updating the information based on uuid so I don't see why that would affect the update.

adriana-alter:
Are you seeing it happen all the time, though (on any branch)?  I agree that it's more important than refreshing the data right away--I just can't reproduce it except in that one weird case...

EdwinGuzman:
Seeing it all the time on any branch: found it happens with multiple libraries with the same contact email.

adriana-alter:
It is a server-side issue unrelated to this branch; I'll give it its own PR.

adriana-alter:
Note: we have established that the behavior involving multiple libraries with the same email address is intentional, not a bug.

PR #10, opened 2019-03-26
Import the Panel component from the library-simplified-reusable-compo…

adriana-alter:
…nents package


src/stylesheets/app.scss
```diff
@@ -1,5 +1,5 @@
 @import "../../node_modules/bootstrap/dist/css/bootstrap.css";
-@import "../../node_modules/opds-web-client/lib/stylesheets/app.scss";
```
EdwinGuzman: Removing `panel.scss` was fine but why did you remove this file? It's still needed... and the new file does not replace this.

src/stylesheets/app.scss
```diff
@@ -1,5 +1,5 @@
...
+@import "../../node_modules/library-simplified-reusable-components/lib/stylesheets/app.scss";
```
EdwinGuzman: Testing this does not work. It doesn't work because that's not the name of the file.
Screen Shot 2019-03-26 at 5 17 29 PM

The name of the compiled css is `reusable-components.css` (https://github.com/NYPL-Simplified/reusable-components/blob/master/webpack.prod.config.js#L25) and it should be retrieved from the `dist` folder since that's the webpack compiled folder.

Also, a better way to import scss or css in a project with webpack is to use `~` since webpack understands to use the `node_modules` folder and you can remove the relative path:
```
@import "~library-simplified-reusable-components/dist/reusable-components.css";
```
(I had wanted to but never got a chance to update all the previous imports).

src/stylesheets/app.scss
adriana-alter: I'll put it back in just in case, but I'm pretty sure this repo isn't actually using anything from it (I think I had just copy/pasted it from circ-web when I was first setting this up)...

src/stylesheets/app.scss
adriana-alter: Ah.  That explains it.  Thank you; good catch!

src/stylesheets/app.scss
EdwinGuzman: The repo is using a few mixins from that repo, unfortunately. If you remove it and try to build, it'll give you that warning.

EdwinGuzman:
Tests passed with latest published npm package. Good to go now.

PR #11, opened 2019-04-03
Update panel

adriana-alter:
The prop to populate the body of the Panel is now named `content` instead of `body`.


EdwinGuzman:
Looks fine. I'll re-check this PR once the reusable component is published to test it directly from npm.

PR #14, opened 2019-04-15
Import button

adriana-alter:
https://jira.nypl.org/browse/SIMPLY-1858


EdwinGuzman:
This looks good but needs some minor updates to fix the conflicts. Also, a jira ticket attached in the description.

I do see this as the active state after clicking a button:
Screen Shot 2019-04-18 at 11 07 16 AM

adriana-alter:
Nice catch!  Fixed via `reusable-components`.

src/components/reusables/Header.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from "react";
 import { LogoutIcon } from "@nypl/dgx-svg-icons";
+import { Button } from "library-simplified-reusable-components";
```
EdwinGuzman: You're not using this anymore.

EdwinGuzman:
Visually this looks good with the latest update. There's one test that fails when running it locally so that should be fixed.

EdwinGuzman:
Looks good. I noticed the styling for the tabs changed and it doesn't look like the tabs on prod...

Screen Shot 2019-04-22 at 11 21 03 AM

adriana-alter:
I will fix the tabs styling when I'm importing the Tabs component; the registry interface isn't being used by anyone outside our team right now, so not worth blocking this PR over.

EdwinGuzman:
That's fine, but if you're only working on one component, the others should not be affected. I would have imagined that the rest stay the same but instead a styling issue was introduced (even if you will update that component later).

adriana-alter:
I know, but it's because the tabs were getting some of their styling information from the button stuff, and now that the button is living elsewhere and has been adapted...  If I do the tabs now, I'll just have to do them all over again, because there will inevitably be similar changes when the tabs component moves.  Not worth doing twice, given how inconsequential the problem is.

PR #15, opened 2019-04-19
Update packages for opds-web-client import

EdwinGuzman:
I believe this is pretty much ready for a local review. No JIRA ticket for this (can't access it) but it's similar to the other tickets for updating packages. For this first pass, I did not update webpack but only updated Typescript, types, and other packages needed to compile the imported `opds-web-client`.

When developing locally, importing the `opds-web-client` using `npm link` makes the tests pass as well as successfully compiling to build the source to run the app.

I tried to test this linking to the github PR branch for `opds-web-client` but that doesn't work because the package itself is in a subfolder of the main repo. I didn't realize that wouldn't work. So now I think to _really_ test and make this work, the `opds-web-client` should be published first and then the travis tests will pass.


package.json
```diff
@@ -21,7 +21,8 @@
     "@nypl/dgx-svg-icons": "^0.3.9",
     "bootstrap": "^3.3.6",
     "library-simplified-reusable-components": "1.3.0",
-    "opds-web-client": "0.1.27",
+    "opds-web-client": "0.2.0",
```
EdwinGuzman: Assuming new version will be `0.2.0`.

src/__tests__/actions-test.ts
```diff
@@ -1,5 +1,6 @@
 import { expect } from "chai";
 import { stub } from "sinon";
+const fetchMock =  require("fetch-mock");
```
EdwinGuzman: This file had the biggest change due to this new package for mocking requests.

src/__tests__/actions-test.ts
```diff
@@ -30,92 +31,92 @@ describe("actions", () => {
...
-      }));
-      fetch = fetchMock;
+      });
+      fetchMock
+        .mock(url, responseText);
 
       await actions.postForm(type, url, formData, "DELETE")(dispatch);
+      let fetchArgs = fetchMock.calls();
+
       expect(dispatch.callCount).to.equal(3);
```
EdwinGuzman: Hmm does a delete request need three calls? (This was pre-PR).

src/__tests__/actions-test.ts
```diff
@@ -30,92 +31,92 @@ describe("actions", () => {
...
 
     it("dispatches failure on bad response", async () => {
       const dispatch = stub();
-      const jsonResponse = new Promise(resolve => resolve({ detail: "problem detail" }));
-      const fetchMock = stub().returns(new Promise<any>(resolve => {
-        resolve({ status: 500, json: () => jsonResponse });
-      }));
-      fetch = fetchMock;
+      fetchMock
+        .mock(url, { status: 500, body: { detail: "problem detail" } });
```
EdwinGuzman: What bothers me about this module is that you can't add any values in the `json` prop. Adding it in the body returns what is needed but it doesn't seem right to me. Following the if/else in the function is still the same as before.

src/__tests__/actions-test.ts
```diff
@@ -151,10 +149,8 @@ describe("actions", () => {
 
     it("dispatches failure on no response", async () => {
       const dispatch = stub();
-      const fetchMock = stub().returns(new Promise<any>((resolve, reject) => {
-        reject({ message: "test error" });
-      }));
-      fetch = fetchMock;
+      fetchMock
+        .mock(url, () => { throw { message: "test error" }; });
```
EdwinGuzman: This took forever to find. It also seems difficult to throw an error (and have no response) but this is how to do it.

src/__tests__/actions-test.ts
```diff
@@ -305,71 +300,81 @@ describe("actions", () => {
...
-      }));
-      fetch = fetchMock;
+      fetchMock
+        .mock(registrationUrl, "done");
 
       await actions.editStages(formData)(dispatch);
-      expect(dispatch.callCount).to.equal(2);
+      let fetchArgs = fetchMock.calls();
+
+      expect(dispatch.callCount).to.equal(3);
```
EdwinGuzman: This had to be updated to `3` but the title of this test is `request and success`, not `request and success and load`. This also seems wrong to me... Does posting data to edit a stage also return data to load?

For the function itself, in `postForm` there was a `response.text()` function but it did not resolve. Since it did not resolve, the tests kept timing out. I'm not sure if it's right for this mocking function to assume that there is value/function for `.text()`.

src/__tests__/actions-test.ts
```diff
@@ -30,92 +31,92 @@ describe("actions", () => {
...
-      expect(fetchMock.args[0][1].headers).to.deep.equal(expectedHeaders);
-      expect(fetchMock.args[0][1].body).to.equal(formData);
+      expect(fetchArgs[0][1].headers).to.deep.equal(expectedHeaders);
+      expect(fetchArgs[0][1].body).to.equal(formData);
     });
 
     it("dispatches a DELETE request", async () => {
       const dispatch = stub();
-      const responseText = stub().returns(new Promise<string>((resolve) => {
+      const responseText = new Promise<string>((resolve) => {
```
aslagle: Why is this a promise when the previous one is just text and the one before that is a stub? They were more similar before.

src/__tests__/actions-test.ts
aslagle: This seems fine to me. Before we had to mock the `json` method on the response since it wasn't a real response, but maybe the new library is creating a more realistic response already.

src/__tests__/actions-test.ts
EdwinGuzman: I was testing all versions out and forgot to just choose one. Will update.

src/__tests__/actions-test.ts
aslagle: I don't know what's supposed to happen, but it does seem questionable that you had to change this.

src/__tests__/actions-test.ts
```diff
@@ -305,71 +296,81 @@ describe("actions", () => {
...
-      }));
-      fetch = fetchMock;
+      fetchMock
+        .mock(emailUrl, "done");
 
       await actions.validateEmail(formData)(dispatch);
-      expect(dispatch.callCount).to.equal(2);
+      let fetchArgs = fetchMock.calls();
+
+      expect(dispatch.callCount).to.equal(3);
```
EdwinGuzman: @adriana-alter these last three set of tests should be calling `dispatch` three times instead of two. It seems that only the failing cases call the dispatch twice. Does this seem right to you? (So then it's request, success, load).

adriana-alter: No, it seems that certain types of (not-failing) cases are in fact supposed to call the dispatch twice; depends on what the action entails.  This action doesn't load anything; it's just a post request.  There are several tests in `circulation-web` that conform to this pattern; [an example](https://github.com/NYPL-Simplified/circulation-web/blob/master/src/__tests__/actions-test.ts#L301).

EdwinGuzman: For the `editBook` action you linked, the server returns `Response("", 200)`. For this test of `validateEmail`, the server returns a dictionary. Both `postForm` functions call the `dispatch` three times (it calls "load" in the `.text()` if/else branch). I'm not sure why the test you linked is only saying it called it twice. I tested this one and it does get called three times in the app, and what gets loaded the dictionary that the server sends. So this, along with the other updates to the dispatch being called three times seems right to me. (@aslagle ).

src/__tests__/actions-test.ts
EdwinGuzman: I'm not sure why it was `2` before but the server returns the `uuid` after the success dispatch in the `response.text()` method. So then the dispatch calls `.load()`. `3` seems to be the right number of calls for this.

EdwinGuzman:
@adriana-alter can you do a final review by installing the new `opds-web-client` package and making sure the app works locally? Thanks.

adriana-alter:
Everything looks good to me!

PR #16, opened 2019-04-26
Search library list

adriana-alter:
[New search feature](https://jira.nypl.org/browse/SIMPLY-1887) for looking up libraries by name.  (The relevant server-side changes have already been merged into `master` of `library_registry`.)


EdwinGuzman:
Haven't reviewed code yet, but from briefly looking at the code, this would have been a good candidate for refinement/talk about implementation beforehand. It's fine that you're making a post request but it would have been nice to make a dynamic search where you can type and the library you want appears as the others don't appear. So, a fuzzy-like search or a typeahead. It could've by searching through the store's libraries data. The post request is good for the no-js scenario but we don't talk much about that for these apps.

src/actions.ts
```diff
@@ -148,6 +149,16 @@ export default class ActionCreator extends BaseActionCreator {
     return this.fetchJSON<LibraryData>(ActionCreator.GET_ONE_LIBRARY, url).bind(this);
   }
 
+  search(data: FormData) {
+    let url = "/admin/libraries/search_by_name";
+    let result = this.postForm(ActionCreator.GET_ONE_LIBRARY, url, data, "POST", "", "JSON").bind(this);
+    return result;
+  }
+
+  clearSearch() {
```
EdwinGuzman: I don't get what this is doing and it's also not attached to any reducers.

src/components/LibrariesList.tsx
```diff
@@ -16,46 +23,80 @@ export interface LibrariesListOwnProps {
...
-          ) :
-          <span className="page-header">There are no libraries in this registry yet.</span>
+          )
         }
-      </ul>
-    );
+        </ul>
+      );
+    }
+    else {
```
EdwinGuzman: This should be on the previous line.

adriana-alter:
Updated with new functionality: rather than scrolling to one search result, the list now gets filtered to display only the search results (and hide all of the other list items).  Note: the motivation for this is so that we will eventually be able to display multiple search results, but the server code does not yet support this functionality; for now, if you want to test how displaying multiple results, you have to hard-code the `search_by_name` method in `library_registry/controller.py` to return a specific array of libraries rather than actually running the search.

src/components/LibrariesListItem.tsx
```diff
@@ -9,6 +9,7 @@ import { GenericWedgeIcon } from "@nypl/dgx-svg-icons";
 export interface LibrariesListItemProps {
   library: LibraryData;
   store: Store<State>;
+  current?: boolean;
```
EdwinGuzman: This prop is not being set for `LibrariesListItem` so it doesn't look like this PR is needed https://github.com/NYPL-Simplified/reusable-components/pull/12/files

adriana-alter: It is not needed.  Will update description.

src/components/LibrariesList.tsx
```diff
@@ -16,34 +17,65 @@ export interface LibrariesListOwnProps {
...
+      return <span className="page-header">There are no libraries in this registry yet.</span>;
+    }
   }
 
   componentWillMount() {
     this.props.fetchData();
   }
 
+  async search(data: FormData) {
+    console.log((data as any).get("name"));
```
EdwinGuzman: Console statement.

src/components/LibrariesList.tsx
```diff
@@ -56,9 +56,8 @@ export class LibrariesList extends React.Component<LibrariesListProps, Libraries
         }
         </ul>
       );
-    }
-    else {
-      return <span className="page-header">There are no libraries in this registry yet.</span>;
+    } else {
+        return <span className="page-header">There are no libraries in this registry yet.</span>;
```
EdwinGuzman: This should still stay at the 2 spaces indentation though.

adriana-alter: That looks SO WRONG to me.  Weird.

src/components/SearchForm.tsx
```diff
@@ -0,0 +1,55 @@
...
+    super(props);
+    this.updateSearchTerm = this.updateSearchTerm.bind(this);
+    this.state = { searchTerm: "" };
+  }
+
+  render(): JSX.Element {
+    let form = (
+      <Form
+        onSubmit={this.props.search}
+        content={<Input name={this.props.inputName} callback={this.updateSearchTerm} />}
```
EdwinGuzman: Make the input it's own variable.

src/stylesheets/global.scss
```diff
@@ -24,3 +24,7 @@
     @include basic-focus;
   }
 }
+
+html {
+  scroll-behavior: smooth;
```
EdwinGuzman: Why do you need this?

adriana-alter: It was for the feature's original behavior, where the list would automatically scroll down to the search result; don't need it anymore.

EdwinGuzman:
I haven't reviewed the tests just yet.

I know the user base for this is just a handful, but since we are trying to build reusable components and build with the correct patterns, we should follow it for Search. One more reason why having a discussion on this beforehand would have helped.

1) If you enter a keyword and no library matches it, the error message `There are no libraries in this registry yet` doesn't make sense.
2) When there are results, you lose the keyword you searched by and that's useful to have.
3) To go along with (2), a pattern we use for search is to give users feedback on their results. From the screenshot below, you'd want to let them know that what they are seeing are "results for X" (you don't need numbers for this unless if you want to). This is where you can also put an error message saying that there are no results (second screenshot).

Screen Shot 2019-04-29 at 12 25 40 PM

Screen Shot 2019-04-29 at 12 27 33 PM

4) From the screenshot above, try to focus on the text that appears that gives feedback to users (in the screenshot it focuses on possible filters but there are none here). It's useful for screenreaders and anyone tabbing through.
5) The `clear search` button isn't aligned and might look better next to the search button.

Screen Shot 2019-04-29 at 12 24 16 PM

adriana-alter:
I completely agree about the `clear search` button needing to be aligned/moved, but that needs to happen in the reusable components repo, not here.

Will work on implementing the additional functionality!

src/actions.ts
```diff
@@ -148,6 +148,12 @@ export default class ActionCreator extends BaseActionCreator {
     return this.fetchJSON<LibraryData>(ActionCreator.GET_ONE_LIBRARY, url).bind(this);
   }
 
+  search(data: FormData) {
+    let url = "/admin/libraries/search_details";
+    let result = this.postForm(ActionCreator.GET_ALL_LIBRARIES, url, data, "POST", "", "JSON").bind(this);
```
EdwinGuzman: Does it make sense to still call the same `GET_ALL_LIBRARIES` action when attempting a search?

src/components/LibrariesList.tsx
```diff
@@ -16,26 +17,53 @@ export interface LibrariesListOwnProps {
...
-          this.props.libraries.libraries.map(library =>
-            <LibrariesListItem
-              key={library.uuid}
-              library={library}
-              store={this.props.store}
-            />
-          ) :
-          <span className="page-header">There are no libraries in this registry yet.</span>
-        }
+        {form}
```
EdwinGuzman: A couple of semantic updates here: the `SearchForm` isn't a list item but a `Panel`. It should not belong in the `ul`, and maybe the error message about no libraries in the list should also be a `li` since the parent is a `ul`.

src/components/LibrariesListItem.tsx
EdwinGuzman: This is still here as well as the `openByDefault` prop in `Panel` below.

src/components/SearchForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+    );
+
+    let form = (
+      <Form
+        onSubmit={this.props.search}
+        content={input}
+        buttonContent={<span>Search <SearchIcon /></span>}
+        className="inline"
+        disableButton={!this.state.searchTerm.length}
+        successText={this.message()["success"]}
```
EdwinGuzman: Make `this.message()` into a variable, that way you don't call the function twice for each separate value.

src/components/LibrariesList.tsx
EdwinGuzman: The library `Panel`s themselves in the ul aren't list items so maybe wrap them around an `li`.

src/components/__tests__/LibrariesListItem-test.tsx
```diff
@@ -75,5 +76,11 @@ describe("LibrariesListItem", () => {
       expect(wrapper.state().color).to.equal("success");
       expect(wrapper.find(".panel-success").length).to.equal(1);
     });
+    it("should optionally be open by default", () => {
```
EdwinGuzman: Don't think you need these anymore.

src/components/__tests__/SearchForm-test.tsx
```diff
@@ -0,0 +1,103 @@
+import { expect } from "chai";
+import * as Sinon from "sinon";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+
+import SearchForm from "../SearchForm";
+import Form from "../reusables/Form";
+
+describe("SearchForm", () => {
+  let wrapper: Enzyme.CommonWrapper<any, any, {}>;
```
EdwinGuzman: Instead of `any`, does `{}` work?

src/components/reusables/Input.tsx
```diff
@@ -3,15 +3,16 @@ import * as React from "react";
...
+  callback?: (e: any) => any;
 }
 
 export default class Input extends React.Component<InputProps, {}> {
   render(): JSX.Element {
     return(
       <div className="form-group">
         <label htmlFor={this.props.name}>{this.props.label}</label>
-        <input className="form-control" type={this.props.type || "text"} name={this.props.name} id={this.props.name} />
+        <input onChange={this.props.callback ? this.props.callback : null} ref="input" className="form-control" type={this.props.type || "text"} name={this.props.name} id={this.props.name} />
```
EdwinGuzman: If these get too long, break each prop into individual lines.

src/stylesheets/list.scss
```diff
@@ -3,9 +3,16 @@
...
   list-style: none;
+  position: relative;
 
   hr {
     margin: 2em auto;
     border-color: $blue-dark;
   }
+
+  .panel-info {
+    position: sticky;
```
EdwinGuzman: Unfortunately, I don't think we use `sticky` because it's not accessible.

src/components/LibrariesList.tsx
```diff
@@ -1,67 +1,27 @@
...
-              key={library.uuid}
-              library={library}
-              store={this.props.store}
-            />
-          ) :
-          <span className="page-header">There are no libraries in this registry yet.</span>
-        }
-      </ul>
+    return (
+          this.props.libraries ?
```
EdwinGuzman: The indentation for this is very far.

src/components/LibrariesPage.tsx
```diff
@@ -0,0 +1,98 @@
...
+export class LibrariesPage extends React.Component<LibrariesPageProps, LibrariesPageState> {
+  constructor(props: LibrariesPageProps) {
+    super(props);
+    this.search = this.search.bind(this);
+    this.clear = this.clear.bind(this);
+    this.state = { showAll: true, searchTerm: "" };
+  }
+
+  render(): JSX.Element {
+    let hasLibraries = (this.props.libraries && this.props.libraries.libraries && this.props.libraries.libraries.length > 0);
```
EdwinGuzman: No needed but you can just do `this.props.libraries.libraries.length` without having to check if it's greater than 0.

src/components/LibrariesPage.tsx
```diff
@@ -0,0 +1,98 @@
...
+    let allLibraries = hasLibraries && this.props.libraries.libraries;
+    let results = !this.state.showAll && this.props.results && this.props.results.libraries;
```
EdwinGuzman: This makes sense but reading it is difficult because it's a complicated set of logic statements. Can you write some comments? I would say that if someone else reading it may have trouble breaking it down easily, comments or more separate steps would be helpful.

src/components/SearchForm.tsx
```diff
@@ -0,0 +1,84 @@
...
+  message(): {} {
+    let message = {};
+    if (!this.props.term) {
+      return message;
+    }
+    if (this.props.resultsCount) {
+      let resultsNumber = `${this.props.resultsCount} ${this.props.resultsCount > 1 ? "results" : "result"}`;
+      message["success"] = `Displaying ${resultsNumber} for ${this.props.term}:`;
+    }
+    else {
```
EdwinGuzman: Should go on the previous line.

src/components/__tests__/LibrariesList-test.tsx
```diff
@@ -43,56 +43,49 @@ describe("LibrariesList", () => {
...
+      let libraryItems = wrapper.find(".panel").not(".panel-info");
       expect(libraryItems.length).to.equal(2);
     });
 
     it("should render each library's  name and short name", () => {
-      let lib1 = wrapper.find(".panel").at(0);
+      let lib1 = wrapper.find(".panel").not(".panel-info").at(0);
       expect(lib1.text()).to.contain("Test Library 1 (lib1)");
-      let lib2 = wrapper.find(".panel").at(1);
+      let lib2 = wrapper.find(".panel").not(".panel-info").at(1);
```
EdwinGuzman: The `.not` negation to get an element is weird. Is there a way to get it without having to call `not`? But not a big deal.

src/components/LibrariesList.tsx
adriana-alter: I tried indenting it a few different ways, and they all look wrong to me; do you have a preference?

EdwinGuzman:
👍  looks good. One final issue: when you clear the search, the search keyword stays in the input.

PR #17, opened 2019-05-08
Update webpack4

EdwinGuzman:
I didn't get a chance to update the Registry Admin to webpack 4 because there were a lot of errors from the production build, but not the dev build. One issue was solved by using uglify, fine.

But then I kept getting issues with some tests crashing the build. I tried to exclude them from the build (which should be done anyway) but some tests still crash it. It turns out that when you delete the `src/components/reusable/__tests__/` folder the build for production doesn't crash and works well. This is very confusing to me because all those tests don't look any different than the other tests.

I'm making this a draft but if you can take a look (maybe a fresh set of eyes will see something off) and if you see anything weird, let me know. Eventually we will be moving all those components and tests to the Reusable Components repo so I'm not very concern about this current version. Currently, I'm excluding the `__tests__` folders (but for whatever reason `src/components/reusable/__tests__/` _still_ doesn't get excluded!) but just ignoring all the warnings from webpack. The dev and prod build work.

If you're okay with moving to webpack 4 and ignoring the warnings (we can remove the warnings from the terminal output), I can move this from draft to ready.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v1.1.2
+#### Updated
+- Updating to Webpack 4.
+
 ### v1.1.1
 #### Added
 - Implemented a search feature for the list of libraries
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v1.1.1
+#### Added
+- Implemented a search feature for the list of libraries
+
 ### v1.1.0
 #### Updated
 - Updated `typings` to `@types`, updated Typescript, and updated unit tests with fetch-mock. These updates are needed to import the updated `opds-web-client` v0.2.0 package.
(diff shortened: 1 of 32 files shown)


adriana-alter:
My only guess would be that it has something to do with where the files live (although I can't imagine why that would be a problem!); what if you just get rid of `reusables/__tests__` and move those tests in with the ones for the other components instead?

If that still doesn't work, then--given that, as you pointed out, this is just a temporary issue because these tests are going to move to a different repo anyway--I'm fine with ignoring the warnings.

EdwinGuzman:
Yep, it's where they live, for whatever reason. If I move the folder up a level, the build doesn't crash. I kinda don't want to do that, however, but it is an option.

Also, those tests weren't caught by the linter so I updated that command and fixed a few issues.

adriana-alter:
It definitely seems worth moving the folder up a level; those tests won't live in this repo soon anyway.

EdwinGuzman:
Alright, moved tests up a folder with the other tests under `components`. Temporary moved but for whatever reason that makes it build without issues.

adriana-alter:
Looks good to me!

PR #20, opened 2019-05-30
Show production libraries

adriana-alter:
Screen Shot 2019-05-30 at 1 02 43 PM
Screen Shot 2019-05-30 at 1 02 53 PM

The Toggle implementation is in accordance with the [ARIA guidelines](https://www.w3.org/TR/wai-aria-practices/#button), which says that toggles should 1) be `button` elements (not checkboxes or radios) with the `aria-pressed` attribute set, and 2) respond to the space and enter keys, but not to the arrow keys.

The Toggle component can accept anything as a callback.  In most cases, I imagine it will get a function that filters existing data rather than hitting an endpoint.  But I am still VERY firmly convinced that, in this specific case, hitting the endpoint makes more sense, because:

-- The default behavior is to show only the production libraries (by hitting the `/libraries` endpoint). 
 Very occasionally, a user will need additional information, i.e. need the longer list obtained from the `/libraries/qa` endpoint.  Conceptually, getting the QA list == making a request for more information, not applying a filter to existing information.
-- Writing a function to filter the QA list on the front end is redundant, given that `/libraries` is already giving us a way to do exactly that.
-- Given that the server defines separate QA and production endpoints and defaults to providing the data from the production endpoint unless otherwise specified (in general, not just for this), it's inconsistent and counterintuitive to have the front end implement the opposite pattern (i.e. only ever use the QA endpoint).  The QA view--which represents a specialized mode which will presumably, as more people start using this interface, only be available to users with elevated permissions--shouldn't be the default.
-- From a performance and scalability standpoint--which will become increasingly relevant as we get more libraries!--it makes no sense to initially load a huge amount of unnecessary, unwanted data and then filter it out, rather than just loading the data that we want in the first place.  (It would be one thing if this were a filter with multiple options, but it's not--it's just an on/off switch, which will usually be left in the off position.)  In fact, in production, this interface had been taking a while to load the list of libraries; having it default to only getting the production libraries has fixed that.

Thoughts?


EdwinGuzman:
That's great that you found that the button can be used as a toggle and it has ARIA guidelines. There are other ways to do it and this seems like a good option. I'd ask Willa to also take a look.

As for the endpoints, it does makes sense to only show the production libraries by default and then load the rest of the libraries if they are needed. I don't agree that you should be making an API call every time you switch the toggle button, regardless if it is called often or not. It's not a good pattern and doesn't make sense to me to have one action for each API call, especially if you are making the same request multiple times without any caching. So even if a user will not need the qa libraries, the architecture in place doesn't make sense to me.

This implementation is an easy way to do this but I don't like the option to be able to make an API call for each click, especially if the data is not changing between toggles. At this point you are just repopulating the store with the same data and just adding/removing new objects while the majority stays the same. You could probably make the default be the production libraries, if the toggle button is hit then fetch the rest of the libraries and update the store with any new library that is not already in the store. After that, I'd rather have a front-end toggle to filter out the libraries based on production status rather than make any more API calls. In this case you are only ever making two calls at most.

I am not convinced right now. You can ask Amy or anyone else to join the discussion and see if there is a different approach or why this pattern is fine.

adriana-alter:
Okay, new pattern:

1) On initial load: server call to get just the production libraries.
2) First time you switch to QA mode: server call to get the QA libraries.
3) On all subsequent toggles, no server call; just use the already-stored values.

EdwinGuzman:
I don't think the animation on the "QA" label button thing makes much sense when it is turning to white. There's also no initial indication that the initial state is production libraries and I think having that stated explicitly would be better (and perhaps toggling the label to say QA/production would be better). I knew what to expect from this update but when I first loaded the page I was still confused.

I don't think each object is being updated correctly:
* If you are on the production list and make a library go to the "testing" stage, when you switch to the QA list and then back to production, that library is still on the production list and the "testing" stage values that were just updated have been reverted to the previous values.
* The same occurs when you go from the QA list and switch a library to the "production" stage. The library does not appear in the production list when you switch, and when you switch back to the QA list, the library has its previous values.

But yes, it is only making two network changes now. Code-wise, which version seems better to you?

EdwinGuzman:
One pattern we've done before and I've seen elsewhere is have two label on each side of the toggle button and highlight which one is active. Maybe that could work?

adriana-alter:
Hmmm.  The toggle is in the off position when you load the page.  I don't think it's a good idea to toggle the label; 1) given that this is just an on/off switch, it makes sense (at least to me!) to conceptualize it as "we are toggling QA mode on and off" rather than "we are toggling between production and QA", and 2) the generally accepted expectations for how toggles work seem to be more along the lines of what we currently have.  [Here's an article](https://www.nngroup.com/articles/toggle-switch-guidelines/) that I found helpful when I was putting this together.  Note that, in all their examples, the label always has the same text; it's just the toggle itself that changes.  (Having two labels seems more radio-button-y.) . All that being said, though, I agree that this is potentially confusing; that was the impetus for doing the animation on the label.  I'll see whether I can come up with something different/additional!

That is...quite a big problem re: updating the objects, yes.  Good catch!  I will investigate.

EdwinGuzman:
That's a good article and good site in general. How about you make it more descriptive like in some of their examples such as "show qa libraries" instead of relying on the animation. The animation is nice but I don't think it adds much and those examples also don't animate the labels.

src/components/LibrariesPage.tsx
```diff
@@ -76,14 +71,64 @@ export class LibrariesPage extends React.Component<LibrariesPageProps, Libraries
...
+    // variable remain undefined; this will generate a "no libraries in this registry" message.
+    return (filteredResults || allResults || noResults) || filteredLibraries || allLibraries;
+  }
+
+  async toggleQA(showQA: boolean) {
+    this.setState({ "qa": showQA });
+    let hasAlreadyLoadedQA: boolean = this.props.libraries.libraries && !(this.props.libraries.libraries.every(l => l.stages.registry_stage === "production"));
+
+    if (!hasAlreadyLoadedQA) {
+      let fetch = showQA ? this.props.fetchQA : this.props.fetchData;
```
EdwinGuzman: If you initially called `fetchData` for the production libraries and only want to call it once, do you need this here?

src/components/LibrariesPage.tsx
```diff
@@ -76,14 +71,64 @@ export class LibrariesPage extends React.Component<LibrariesPageProps, Libraries
...
+      await fetch();
+    }
+    if (showQA && this.state.searchTerm && !this.props.results) {
```
EdwinGuzman: Nice.

src/components/LibrariesPage.tsx
adriana-alter: Pretty sure I do not; good catch!

src/components/LibrariesPage.tsx
```diff
@@ -76,14 +71,64 @@ export class LibrariesPage extends React.Component<LibrariesPageProps, Libraries
...
   async search(data: FormData) {
     this.setState({ showAll: false, searchTerm: (data.get("name") as string) });
     await this.props.search(data);
   }
 
   async clear() {
     this.setState({ showAll: true, searchTerm: "" });
-    await this.props.fetchData();
+    let fetch = this.state.qa ? this.props.fetchQA : this.props.fetchData;
+    await fetch();
```
EdwinGuzman: So when you search for a term, it's making a new API call and clearing the store of the libraries. That's why there's a need to refetch the QA or production libraries endpoint again? I forgot that this made an API call when clearing the search, I thought it was just un-filtering the data set to show all the libraries again.

src/components/__tests__/LibrariesPage-test.tsx
```diff
@@ -102,32 +175,57 @@ describe("LibrariesPage", () => {
...
+
+      results = wrapper.find(".panel-warning");
+      expect(results.length).to.equal(1);
+      expect(results.find(".panel-title").text()).to.equal("QA Library (qa)");
+    });
+
+    it("should let the user check QA if the production search was unsuccessful", async () => {
+      expect(search.callCount).to.equal(0);
+      wrapper.setState({ searchTerm: "QA Library"});
+      wrapper.instance().toggleQA(true);
```
EdwinGuzman: `toggleQA` is asynchronous but `await` didn't work for this (and so the `pause` function)?

src/components/LibrariesPage.tsx
adriana-alter: Yes.  Do you think it would be better to just have the search happen on the front end (i.e. be a filter)?  I did it the way it is because there was already a search method in the server code, so it seemed clunky to duplicate the existing functionality on the front end.  But on the other hand, moving the search to the front end would make some things much easier.

EdwinGuzman: Given the API endpoint that does the search for you and not wanting to duplicate functionality, I think it's okay. Everything on the front-end has its pros and cons and for this, relying on the server code is fine. You'd have multiple calls to search and clear (and then possibly switch lists):
Screen Shot 2019-06-04 at 11 40 32 AM

It works fine now and I would say leave this part as is. This is not really part of the PR anyway, but if you want to rethink the architecture for this later that'd be a different convo.

src/components/reusables/Toggle.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from "react";
```
EdwinGuzman: 👍 
Good thing you made this its own form.

src/components/reusables/Toggle.tsx
```diff
@@ -21,8 +19,12 @@ export default class Toggle extends React.Component<ToggleOwnProps, ToggleState>
     this.state = { on: this.props.initialOn || false };
   }
 
+  componentWillReceiveProps(newProps: ToggleOwnProps) {
+    this.setState({ on: newProps.initialOn });
+  }
+
   render(): JSX.Element {
-    let label = this.props.label || "On";
+    let label = this.props.label ? `${this.props.label}: ${this.state.on ? "On" : "Off"}` : "On";
```
EdwinGuzman: So if you don't pass in a label, it'll always show the default of `on`?

package.json
```diff
@@ -20,7 +20,7 @@
   "dependencies": {
     "@nypl/dgx-svg-icons": "^0.3.9",
     "bootstrap": "^3.3.6",
-    "library-simplified-reusable-components": "1.3.7",
+    "library-simplified-reusable-components": "^1.3.7",
```
EdwinGuzman: Why the caret?

EdwinGuzman: I don't like using carets because they can create dependencies issues if any new packages.

src/components/reusables/Toggle.tsx
adriana-alter: Yes.  Having the whole thing toggle to "Off" doesn't seem to match the patterns with making toggles; you're supposed to be toggling whether or not it's on, not whether it's on or off.  (I am aware that this is an extremely subtle difference; if you'd rather it toggle to "Off", I do not have strong objections or anything...)

package.json
adriana-alter: It was suddenly breaking otherwise.  No idea why.

adriana-alter:
I think I tried to do that test, and there was some issue with it.  Will try again, though!

package.json
adriana-alter: I know, but...this seems like a reasonable exception, given that it's our package.  Does it work for you without the caret?

src/components/reusables/Toggle.tsx
EdwinGuzman: Yes, having "off" doesn't match the patterns. I'm okay with this.

package.json
EdwinGuzman: It's not breaking and it should be fine with our package, but in general for other packages we don't use. That reminds me, it seems an updated package.json should be committed (a new version comes up when I run `npm install`).

src/components/reusables/Toggle.tsx
adriana-alter: What's "this"--okay with not matching the patterns, or okay with the current implementation?

EdwinGuzman: With the implementation.

package.json
adriana-alter: And...it's working without the caret again.  It's a mystery.  That's...nice.

EdwinGuzman: Maybe an issue with the updated package-lock file.

PR #21, opened 2019-05-31
Add "Areas" tab in library detail view

adriana-alter:
Screen Shot 2019-05-31 at 4 29 20 PM

Goes with [this server branch](https://github.com/NYPL-Simplified/library_registry/pull/119).

The purpose of this branch is to add an optional "Areas" tab to the tabs in the library detail view (optional because if the library doesn't have any data about areas, there's no point displaying a blank tab; better to just leave the tab off).

While trying to update all of the tests that were suddenly complaining because their sample library data didn't reflect the change to the library interface, I got fed up and created a `library-data` file from which all of the tests can import sample library objects.  (It's the same library objects that the tests were already using; it's just that now they live in their own file rather than being copy-pasted into every single test.  Reusable stuff for the win!!!)  There is room for several other test improvements along similar lines, but this isn't the branch for them.

### DIFF
--- a/src/components/LibraryDetailPage.tsx
+++ b/src/components/LibraryDetailPage.tsx
@@ -99,9 +99,10 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
       "Contact & URLs": "urls_and_contact",
       "Areas": "areas"
     };
-    Object.entries(categories).filter(([k, v]) => library[v]).forEach(([k, v]) => {
+    Object.entries(categories).forEach(([k, v]) => {
+    // Object.entries(categories).filter(([k, v]) => library[v]).forEach(([k, v]) => {
       let category = library[v];
-      let categoryItems: any[] = Object.values(category);
+      let categoryItems: (string | string[])[] = Object.values(category);
       // If there are no meaningful items in this category--e.g. it's an Areas category in which
       // both values are empty arrays--then don't bother making a blank tab for it.
       let hasItems = categoryItems.some(x => x && x.length > 0);
--- a/src/components/LibraryDetailPage.tsx
+++ b/src/components/LibraryDetailPage.tsx
@@ -46,10 +46,14 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
   }
 
   renderItems(category: {[key: string]: string}): JSX.Element {
+    // If the value is an array--e.g. of focus or service areas--format it as a semicolon-separated list.
+    let formatValue = (value: string | string[]) => {
+      return (Array.isArray(value) ? value.join("; ") : `${value}`);
+    };
 
     // Only create LibraryDetailItems for fields which actually have a value.
-    let fields = Object.keys(category).filter(label => category[label]).map(label =>
-      <LibraryDetailItem key={label} label={label} value={`${category[label]}`} />
+    let fields = Object.keys(category).filter(label => category[label] && category[label].length).map(label =>
+      <LibraryDetailItem key={label} label={label} value={formatValue(category[label])} />
     );
 
     return (
@@ -92,11 +96,16 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
 
     const categories = {
       "Basic Information": "basic_info",
-      "Contact & URLs": "urls_and_contact"
+      "Contact & URLs": "urls_and_contact",
+      "Areas": "areas"
     };
-
-    Object.keys(categories).forEach(tabName => {
-      tabItems[tabName] = this.renderItems(library[categories[tabName]]);
+    Object.entries(categories).filter(([k, v]) => library[v]).forEach(([k, v]) => {
+      let category = library[v];
+      let categoryItems: any[] = Object.values(category);
+      // If there are no meaningful items in this category--e.g. it's an Areas category in which
+      // both values are empty arrays--then don't bother making a blank tab for it.
+      let hasItems = categoryItems.some(x => x && x.length > 0);
+      hasItems && (tabItems[k] = this.renderItems(category));
     });
 
     return(
--- a/src/components/__tests__/EmailValidationForm-test.tsx
+++ b/src/components/__tests__/EmailValidationForm-test.tsx
@@ -3,6 +3,7 @@ import * as Sinon from "sinon";
 import * as Enzyme from "enzyme";
 import * as React from "react";
 import buildStore from "../../store";
+import { testLibrary1 } from "./library-data";
 import { EmailValidationForm } from "../EmailValidationForm";
 
 describe("EmailValidationForm", () => {
@@ -12,25 +13,8 @@ describe("EmailValidationForm", () => {
   let fetchLibrary: Sinon.SinonStub;
 
   beforeEach(() => {
-    let library = {
-      uuid: "UUID1",
-      basic_info: {
-        "name": "Test Library 1",
-        "short_name": "lib1",
-        "description": undefined
-      },
-      urls_and_contact: {
-        "authentication_url": "auth1",
-        "contact_email": "email1",
-        "opds_url": "opds1",
-        "web_url": "web1",
-        "validated": "Not validated"
-      },
-      stages: {
-        "library_stage": "production",
-        "registry_stage": "testing"
-      }
-    };
+    let library = {...testLibrary1, ...{urls_and_contact: {...testLibrary1.urls_and_contact, ...{validated: "Not validated"}}}};
+
     store = buildStore();
     validateEmail = Sinon.stub();
     fetchLibrary = Sinon.stub();
--- a/src/components/__tests__/EmailValidationForm-test.tsx
+++ b/src/components/__tests__/EmailValidationForm-test.tsx
@@ -13,7 +13,7 @@ describe("EmailValidationForm", () => {
   let fetchLibrary: Sinon.SinonStub;
 
   beforeEach(() => {
-    const library = {...testLibrary1, ...{urls_and_contact: {...testLibrary1.urls_and_contact, ...{validated: "Not validated"}}}};
+    const library = {...testLibrary1, urls_and_contact: {...testLibrary1.urls_and_contact, validated: "Not validated"}};
 
     store = buildStore();
     validateEmail = Sinon.stub();
--- a/src/components/__tests__/EmailValidationForm-test.tsx
+++ b/src/components/__tests__/EmailValidationForm-test.tsx
@@ -3,7 +3,7 @@ import * as Sinon from "sinon";
 import * as Enzyme from "enzyme";
 import * as React from "react";
 import buildStore from "../../store";
-import { testLibrary1, modify } from "./library-data";
+import { testLibrary1, modifyLibrary } from "./TestUtils";
 import { EmailValidationForm } from "../EmailValidationForm";
 
 describe("EmailValidationForm", () => {
@@ -13,7 +13,7 @@ describe("EmailValidationForm", () => {
   let fetchLibrary: Sinon.SinonStub;
 
   beforeEach(() => {
-    const library = modify(testLibrary1, {"validated": "Not validated"}, "urls_and_contact");
+    const library = modifyLibrary(testLibrary1, {"validated": "Not validated"}, "urls_and_contact");
     store = buildStore();
     validateEmail = Sinon.stub();
     fetchLibrary = Sinon.stub();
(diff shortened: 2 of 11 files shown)


src/components/LibraryDetailPage.tsx
```diff
@@ -92,11 +96,16 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
...
     const categories = {
       "Basic Information": "basic_info",
-      "Contact & URLs": "urls_and_contact"
+      "Contact & URLs": "urls_and_contact",
+      "Areas": "areas"
     };
-
-    Object.keys(categories).forEach(tabName => {
-      tabItems[tabName] = this.renderItems(library[categories[tabName]]);
+    Object.entries(categories).filter(([k, v]) => library[v]).forEach(([k, v]) => {
```
EdwinGuzman: What are you filtering here for?

src/components/LibraryDetailPage.tsx
```diff
@@ -92,11 +96,16 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
...
+      let category = library[v];
+      let categoryItems: any[] = Object.values(category);
```
EdwinGuzman: I think you can be more specific with what type this is.

src/components/__tests__/LibrariesListItem-test.tsx
```diff
@@ -3,29 +3,13 @@ import { expect } from "chai";
...
-      "contact_email": "test_email",
-      "opds_url": "test_opds",
-      "web_url": "test_web"
-    },
-    stages: {
-      "library_stage": "production",
-      "registry_stage": "testing"
-    }
-  };
+  let library = testLibrary1;
```
EdwinGuzman: Make it `const` since it's not being updated.

src/components/__tests__/LibrariesPage-test.tsx
```diff
@@ -4,48 +4,14 @@ import * as Enzyme from "enzyme";
...
-          "opds_url": "opds2",
-          "web_url": "web2"
-        },
-        stages: {
-          "library_stage": "testing",
-          "registry_stage": "cancelled"
-        }
-      }
-    ];
+    let libraries = [testLibrary1, testLibrary2];
```
EdwinGuzman: Same for the rest.

src/components/__tests__/library-data.ts
```diff
@@ -0,0 +1,43 @@
...
+  uuid: "UUID2",
+  basic_info: {
+    "name": "Test Library 2",
+    "short_name": "lib2",
+  },
+  urls_and_contact: {
+    "authentication_url": "auth2",
+    "contact_email": "email2",
+    "opds_url": "opds2",
+    "web_url": "web2"        },
```
EdwinGuzman: Weird placement of brace?

src/interfaces.ts
```diff
@@ -10,13 +10,17 @@ export interface LibraryData {
...
   };
-  urls_and_contact?: {
+  urls_and_contact: {
     authentication_url: string,
     contact_email: string,
     opds_url: string,
     web_url: string,
     validated?: string
   };
+  areas: {
```
EdwinGuzman: I think `areas` can also be optional if both its attributes are also optional.

src/components/LibraryDetailPage.tsx
adriana-alter: I had been filtering to make sure that the library object actually has this category (under the assumption that `library[areas]`--and probably other things as we add more features!--might not exist).  But, looking back at the latest version of the server code, I realize that there's no way for that to happen (if the library doesn't have areas, then `library[areas]` == `{focus: [], service: []}`); will remove the filter!

src/interfaces.ts
adriana-alter: The library will always have an `areas` value (see comment re: filter, up top!).  If its focus/service areas aren't specified, then `library[areas]` will be `{focus: [], service: []}`.  (As I type this, I realize that this means that those attributes aren't actually optional; hmm...).  But `areas` itself is always there.

EdwinGuzman: Yeah, that makes sense. But can the server return `areas: null`?

src/components/LibraryDetailPage.tsx
```diff
@@ -99,9 +99,10 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
       "Contact & URLs": "urls_and_contact",
       "Areas": "areas"
     };
-    Object.entries(categories).filter(([k, v]) => library[v]).forEach(([k, v]) => {
+    Object.entries(categories).forEach(([k, v]) => {
+    // Object.entries(categories).filter(([k, v]) => library[v]).forEach(([k, v]) => {
```
EdwinGuzman: You can remove the commented out code.

src/components/LibraryDetailPage.tsx
```diff
@@ -99,9 +99,10 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
...
       let category = library[v];
-      let categoryItems: any[] = Object.values(category);
+      let categoryItems: (string | string[])[] = Object.values(category);
```
EdwinGuzman: In what case will this return an array of array of strings?

src/interfaces.ts
adriana-alter: No--will always have `focus` and `service`.

src/components/LibraryDetailPage.tsx
adriana-alter: The geographical data.  (Because you can have multiple locations per value.)

src/components/__tests__/EmailValidationForm-test.tsx
```diff
@@ -12,25 +13,8 @@ describe("EmailValidationForm", () => {
...
-        "opds_url": "opds1",
-        "web_url": "web1",
-        "validated": "Not validated"
-      },
-      stages: {
-        "library_stage": "production",
-        "registry_stage": "testing"
-      }
-    };
+    const library = {...testLibrary1, ...{urls_and_contact: {...testLibrary1.urls_and_contact, ...{validated: "Not validated"}}}};
```
EdwinGuzman: You don't need the spread operator for `{urls_and_contact: ...` or `validated: "Not validated"` but the rest are okay. So it would be something like
```
const library = {...testLibrary1, urls_and_contact: { ...testLibrary1.urls_and_contact, validated: "Not validated" }};
```

src/components/__tests__/LibraryDetailPage-test.tsx
```diff
@@ -108,6 +98,17 @@ describe("LibraryDetailPage", () => {
     expect(infoLabels.indexOf("description")).to.equal(-1);
   });
 
+  it("should not display empty tabs", () => {
+    let tabs = wrapper.find(".tab-nav");
+    expect(tabs.length).to.equal(3);
+    expect(wrapper.find(".tab-navs").text()).to.contain("Areas");
+    let libraryWithoutAreas = {...library, ...{areas: {focus: [], service: []}}};
```
EdwinGuzman: I think here you can also remove the spread for `areas`.

EdwinGuzman: You want to use it for objects where you want all the attributes but not necessary for the attribute you want to override.

EdwinGuzman:
I don't necessarily see the advantage of the new `modify` function because in the end it's still doing the spread operation for creating a new object from the base object. I think that should've been a different PR to not mix it in with this PR. `modify` is also a bit too general of a name but it is only currently being used for tests, but if it wasn't then "modify" could mean anything. In this case, the `library-data.ts` test file could use a different name since it's not just data anymore. Typically we use "utils" as a folder or file name so that may help?

You should use the esdoc syntax for the function
```
/**
 * modify()
 * @param ....
 */
```

adriana-alter:
The advantage is that 1) it's much easier to call `modify` from the tests than to type out the whole object-creating process every single time (I kept breaking tests by accidentally mistyping something while doing that), and 2) removing those lines from the tests makes the remaining code more readable.  I'll change the function and file names!

EdwinGuzman:
Hmm sure, if it’s easier for you to read that’s fine. I am finding that, as someone new reading the code, it’s not easier or more readable. I find myself looking back between the function definition and how it’s called in the test file. Just something to consider in general for functions for any new developers who just into the code.

adriana-alter:
Interesting!  It's WAY WAY easier for me to read (and write).  Would more comments help?

EdwinGuzman:
Okay then let’s leave it this way. Oh! Also, I don’t think I saw any tests for the function.

src/components/__tests__/TestUtils.ts
```diff
@@ -0,0 +1,73 @@
...
+    "registry_stage": "cancelled"
+  },
+  areas: {
+    "focus": [],
+    "service": []
+  }
+};
+
+export const modifyLibrary = (baseLibrary: LibraryData, newData: {[key: string]: string | string[]}, category?: string): LibraryData => {
+  /**
```
EdwinGuzman: Ah, you'd want the comment above the function declaration.

src/components/__tests__/TestUtils.ts
```diff
@@ -0,0 +1,73 @@
...
+   * modifyLibrary()
+   * @param {LibraryData} baseLibrary - the library to be modified
+   * @param {{[key: string]: string | string[]}} newData - key-value pairs with which to update the library
+   * @param {string} [category] - the category that the new values should go into.  This isn't necessary if we're
```
EdwinGuzman: I don't think the brackets are necessary.

adriana-alter: The brackets indicate that it's an optional argument (according to the esdoc docs).

EdwinGuzman: Cool, I thought it was something else.

src/components/__tests__/TestUtils.ts
```diff
@@ -0,0 +1,119 @@
...
+/**
+* modifyLibrary()
+* @param {LibraryData} baseLibrary - the library to be modified
+* @param {{[key: string]: string | string[]}} newData - key-value pairs with which to update the library
+* @param {string} [category] - the category that the new values should go into.  This isn't necessary if we're
+modifying the value of a key that already exists in the library; modifyLibrary will figure it out.
+It's only necessary if we're adding a new key.
+*/
+export const modifyLibrary = (baseLibrary: LibraryData, newData: {[key: string]: string | string[]}, category?: string): LibraryData => {
+  let categories = Object.keys(baseLibrary);
```
EdwinGuzman: You're not using this value in this function.

src/components/__tests__/TestUtils.ts
```diff
@@ -0,0 +1,119 @@
...
+  Object.entries(newData).forEach((pair) => {
+    let categoryToUpdate = category || null;
+    let newValue = {};
+    let [key, value] = pair;
+    if (!categoryToUpdate) {
+      let allCategories = Object.keys(baseLibrary);
+      categoryToUpdate = allCategories.find((cat) => Object.keys(baseLibrary[cat]).includes(key));
+    }
+    newValue[key] = value;
+    let updatedCategory = { ...(updatedLibrary[categoryToUpdate] || baseLibrary[categoryToUpdate]), ...newValue };
```
EdwinGuzman: This is confusing. Isn't `updatedLibrary` a copy of `baseLibrary` from above? Neither `updatedLibrary` or `baseLibrary` get modified until you call this here.

EdwinGuzman: I guess for the first iteration it won't matter but the rest of the loop is fine.

EdwinGuzman: Actually, since the `categoryToUpdate` value will always be one of the keys from the `baseLibrary` and `updatedLibrary` is created from it, then can't you only just use `updatedLibrary` here?

EdwinGuzman:
I was thinking that it'd be great if it was recursive and could be use in any object data type but it is kind of specific to the `LibraryData` data type. Maybe you want to write a comment about how it only goes one level deep.

src/components/__tests__/TestUtils.ts
adriana-alter: You can, but I think that makes it a little harder to read.  See how it looks to you...

adriana-alter:
There is only one level for it to go; it's not like any of the values within the categories are objects.  If this lived in a different repo, I would look into making it work with other object data types, but within `registry_admin`, there isn't anything else you would ever have occasion to use it on.

src/components/__tests__/TestUtils.ts
EdwinGuzman: Looks more readable because it makes sense where the value is coming from.

PR #22, opened 2019-06-03
React16

EdwinGuzman:
This fixes [SIMPLY-1963](https://jira.nypl.org/browse/SIMPLY-1963).

* This uses the latest version of the `opds-web-client` which is on React16.
* React is updated to v16, Enzyme to v3.x, and jsdom to v15.
* * jsdom is not a necessary update but I thought it would fix an issue I encountered. Updating did not solve the issue :( (but at least updated the `testHelper` file).
* The new v16 ref API is nice and I tried to use but not for everything. Basically, I wanted to use the `forwardedRef` functional component from react but it's a big change so I made a minor update so we can all get acquainted to it.
* `FormData` with refs and testing does not work in the `Form` component with this new API. I tried to find how/why but didn't get far. It seems to be an issue with JSDOM so hence its update, but that still did not fix it. This is probably the only issue I found that had the exact same problem https://github.com/airbnb/enzyme/issues/1871 (and the reply back was with it being a jsdom issue).
* * To work around this issue, I mocked the `Form` prop search functions that were being passed down in a few tests (more in the comment files).
* Enzyme's `.parent()` doesn't work anymore. The workaround is `.parents().at(0)`... until this is fixed.
* This is similar to the React16 update PR for [circ web](https://github.com/NYPL-Simplified/circulation-web/pull/237).


src/components/SearchForm.tsx
```diff
@@ -30,6 +30,7 @@ export default class SearchForm extends React.Component<SearchFormOwnProps, Sear
   render(): JSX.Element {
     let input = (
       <Input
+        key="form-input"
```
EdwinGuzman: This (and adding the key to Form) fixes a few console error issues.

src/components/__tests__/LibrariesPage-test.tsx
```diff
@@ -7,127 +7,147 @@ import buildStore from "../../store";
...
+    // Clear the search results:
+    wrapper.setState({ showAll: true });
+
+    list = wrapper.find(LibrariesList);
+    expect(list.prop("libraries")).to.equal(wrapper.prop("libraries").libraries);
+
+    // Unsuccessful search:
+    wrapper.setState({ showAll: false });
+    wrapper.setProps({ results: null })
+    ;
```
EdwinGuzman: This is weird and doesn't appear like this in my editor!

src/components/reusables/Form.tsx
```diff
@@ -30,20 +32,16 @@ export default class Form extends React.Component<FormProps, {}> {
...
-      if (this.refs["successMessage"]) {
-        (this.refs["successMessage"] as HTMLElement).focus();
-      }
-    }).catch(err => {
-      if (this.refs["errorMessage"]) {
-        (this.refs["errorMessage"] as HTMLElement).focus();
-      }
-    });
+
+    await this.props.onSubmit(data);
```
EdwinGuzman: No need to use then/catch since the function declared is async.

src/components/reusables/Form.tsx
```diff
@@ -19,6 +19,8 @@ export interface FormProps {
 }
 
 export default class Form extends React.Component<FormProps, {}> {
+  private messageRef = React.createRef<HTMLParagraphElement>();
```
EdwinGuzman: The new way to declare refs. It is passed down to the `p` in line 44 and then referenced as `this.messageRef.current`.

src/components/reusables/Form.tsx
```diff
@@ -30,20 +32,16 @@ export default class Form extends React.Component<FormProps, {}> {
     event.preventDefault();
     let form = (this.refs["form"] as any);
```
EdwinGuzman: But, to go with the `messageRef`, creating a ref for this form is more complicated because its ref is referenced by parent components.

src/components/reusables/Form.tsx
aslagle: To make sure I understand this - using a single ref rather than two refs depending on the message type is unrelated to the new ref creation, it's just separate clean up?

src/components/reusables/Form.tsx
aslagle: Could you explain this more? I don't see why it's more complicated.

src/components/reusables/Form.tsx
EdwinGuzman: Correct, using a single ref is more cleanup.

src/components/reusables/Form.tsx
EdwinGuzman: The ref for the form is just as simple to create as the `messageRef` but that ref gets complicated because it's called by parent components. `SearchForm` imports `Form` and uses the `Form`'s ref to reset it: `(this.refs["form-component"] as any).refs["form"].reset();`. Right now this works but setting it up with the new Ref API doesn't work unless more ref forwarding is added. I tried it and got close to making it work but there were some issues I couldn't easily fix and it was also out of scope. The syntax and extra boilerplate was also tedious to work with.

More of ref forwarding here: https://reactjs.org/docs/forwarding-refs.html

adriana-alter: At some point in the not-too-distant future, I'll be working on making `Form` more easily reusable; I'll make a note to take a look at this then!  (Off the top of my head: `Form` should probably have its own built-in `clear` method anyway, which will get rid of this issue altogether.)

adriana-alter:
WOW the `.parent()` thing is annoying!

But anyway.  This is looking good to me.

src/components/reusables/Form.tsx
EdwinGuzman: Yeah, it makes sense to have the `clear` function in the form itself. I can make a different PR for the functional forwarding ref way of doing it but not sure if it will be helpful or not. I'll just wait for your implementation and see.

EdwinGuzman:
@adriana-alter, should I wait until your two PRs get merged in or do you want to fix the conflicts and update your tests based on this PR? (I don't mind fixing any conflicts so I can merge this in after).

adriana-alter:
Would I need to update my tests as part of those PRs in order to make them pass, or could that happen later?  If the former, I'd rather merge first; there's already a lot going on in my PRs (especially the toggle one), so updating the tests seems very out of scope for them.  (I'm happy to do subsequent PRs to update them, though; not trying to make extra work for you!)

EdwinGuzman:
Yes, you'd need to update the tests to pass in your PRs. It will be more work for me but I don't mind. It's also updating the syntax rather than the tests as a whole. I'll wait.

PR #23, opened 2019-06-14
Import reusable Form

adriana-alter:
Goes with https://github.com/NYPL-Simplified/reusable-components/pull/19


adriana-alter:
Note: all the tests pass no problem in local.  Almost positive they'll pass on here once the new ReusableComponents is published and this branch is using it.

EdwinGuzman:
Is the PR for Reusable Components finished? I'll review once they're both complete.

adriana-alter:
It won't be officially finished until I'm done importing the Form into circ-web; I imagine that process is going to involve making some changes/additions to the Form...  But the stuff relevant to registry_admin is finished, and the storybook stuff is finished.

EdwinGuzman:
Okay, I will wait until they are all done in case you need to make unforeseen changes.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v1.2.1
+#### Updated
+- Integrated the Form component from the [https://github.com/NYPL-Simplified/reusable-components](library-simplified-reusable-components) package
```
EdwinGuzman: `.` at the end.

src/components/SearchForm.tsx
```diff
@@ -79,7 +78,7 @@ export default class SearchForm extends React.Component<SearchFormOwnProps, Sear
 
   clear(): void {
     this.props.clear();
-    (this.refs["form-component"] as any).refs["form"].reset();
+    (this.refs["form-component"] as any).formRef.current.reset();
```
EdwinGuzman: Nice, this seems to work.

EdwinGuzman:
This seems okay but there's this bug when you focus on the clear button:
Screen Shot 2019-06-17 at 3 24 55 PM

adriana-alter:
The bug with the button text color will need to be fixed in ReusableComponents, not here.  (And it's not coming from this branch; I just checked, and unfortunately it is on production too.)

src/components/EmailValidationForm.tsx
```diff
@@ -54,7 +54,7 @@ export class EmailValidationForm extends React.Component<EmailValidationFormProp
 
     return (
         <Form
-          className="validation"
+          className="validation border"
```
EdwinGuzman: It seems that this `border` class appears all the time. It makes more sense to have this be the default and make the dashed border an optional class.

EdwinGuzman:
The log in form isn't centered anymore:
Screen Shot 2019-06-26 at 9 52 58 AM

When you search and no results are found, the error messaged isn't focused. This works fine when a result is found.

src/components/EmailValidationForm.tsx
adriana-alter: The dashed border isn't a real class; I just used it for the Storybook examples, because it's otherwise hard to visually differentiate between the different forms when they're all in a list like that.  I initially had the default be border (and had a `no-border` class), but I changed it because I noticed that, in circ-web, the overwhelming majority of the forms needed to be `no-border`, so I figured that the much much more common use case should be the default...

adriana-alter:
Hmmm.  The log-in form is definitely centered for me.  Open the console and check whether it has the `centered` class?  (I can't imagine why that would possibly be different on your computer, but...)

Screen Shot 2019-06-26 at 10 35 50 AM

adriana-alter:
Good catch re: the search error message; will fix!

adriana-alter:
The issue with the error message in the search form has been resolved by the latest ReusableComponents commit.

src/components/EmailValidationForm.tsx
EdwinGuzman: Sounds fine.

EdwinGuzman:
The error message issue initially works but it's on `componentDidUpdate`. If you start typing into the search bar, each key makes the page focus on the message and is annoying.

EdwinGuzman:
It also seemed better to have the search button right next to the input field instead of being all the way at the right side now.

EdwinGuzman:
Looks good. The "Clear search" button is still appearing as white text on white background when tab focusing on it.

adriana-alter:
It's not from this branch.  There's [already a ticket for it](https://jira.nypl.org/browse/SIMPLY-2058).

EdwinGuzman:
I thought it was fixed already. Then I realized this is the same PR where I initially noticed this. I think it should be part of this PR or a separate PR to be merged into this one.

adriana-alter:
It can't be part of this PR; this is the wrong repo for it.  I'll branch off the current ReusableComponents PR, though!

PR #26, opened 2019-07-12
Updating redux, opds-web-client, and typescript. Removing context from the LogInForm.

EdwinGuzman:
Resolves [SIMPLY-2120](https://jira.nypl.org/browse/SIMPLY-2120). Similar to the circulation-web update, I had to update Typescript as well.

For some reason, the update made a few tests failed when loading the `LogInForm`. Apparently the ContextProvider was not successfully passing down the context and its store to that component. I explicitly passed it down and removed the store context from the `LogInForm` since it pick it up at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v1.2.6
+#### Updated
+- Updated the version of the opds-web-client, Redux, and Typescript.
+
 ### v1.2.5
 #### Updated
 - Updated the minifier configuration options.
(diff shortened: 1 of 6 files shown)


aslagle:
This looks fine to me but I'm not sure about the `LogInForm` thing. I'm not familiar enough with this code to comment without investigating more.

EdwinGuzman:
@adriana-alter

adriana-alter:
No idea what would be causing the problem (particularly given the the `LibrariesListContainer` is still fine...right??), but the solution looks fine to me!

EdwinGuzman:
Yes, `LibrariesListContainer` is working fine. I think it's because of how the routing is set up: `LibrariesListContainer` is rendered through the Router but the `LogInForm` is not. A different PR to manage the routing would be nice. I'll create one. Thanks.

PR #27, opened 2019-07-17
Implement form for PLS ID, and show PLS ID if there is one

adriana-alter:
https://jira.nypl.org/browse/SIMPLY-2124


src/components/LibrariesListItem.tsx
```diff
@@ -54,7 +54,8 @@ export default class LibrariesListItem extends React.Component<LibrariesListItem
   render(): JSX.Element {
     let style = this.state.color;
     let { name, short_name } = this.props.library.basic_info;
-    let headerText = `${name} (${short_name})`;
+    let plsID = this.props.library.basic_info.pls_id ? `: ${this.props.library.basic_info.pls_id}` : "";
```
EdwinGuzman: Is this information important enough to view in the header with the name of the library? It’s confusing enough to see the name and then the short name (a random string of chars) in parenthesis in the header. Then this ID would be there without any label. I don't think having this here helps.

src/components/reusables/Input.tsx
```diff
@@ -22,6 +23,8 @@ export default class Input extends React.Component<InputProps, {}> {
           type={this.props.type || "text"}
           name={this.props.name}
           id={this.props.name}
+          defaultValue={this.props.value}
+          readOnly={false}
```
EdwinGuzman: Hmm is this because of the `defaultValue` prop? I run into issues with the `defaultValue` and `value` props for `input` in React.

src/components/LibraryDetailPage.tsx
```diff
@@ -119,6 +120,8 @@ export class LibraryDetailPage extends React.Component<LibraryDetailPageProps, L
           fetchLibrary={this.props.fetchLibrary}
         />
         <hr></hr>
+        <PlsIDForm store={this.props.store} current={library.basic_info.pls_id} fetchLibrary={this.props.fetchLibrary} uuid={library.uuid} />
+        <hr></hr>
```
EdwinGuzman: Oh, for this I think it’s fine if you just do `<hr />` since it’s a self-closing element, but not a big deal.

src/components/PlsIDForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export class PlsIDForm extends React.Component<PlsIDFormProps, PlsIDState> {
+  constructor(props: PlsIDFormProps) {
+    super(props);
+    this.state = { saved: false };
+    this.submit = this.submit.bind(this);
+  }
+  async submit(data: FormData) {
+    await this.props.plsID(data);
```
EdwinGuzman: Maybe a variable name like `postPlsID` would be helpful. I originally thought it was fetching a pls id.

EdwinGuzman: Also to go along with other verb-y action names.

src/components/PlsIDForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+    await this.props.fetchLibrary(this.props.uuid);
+    this.setState({ saved: !this.props.error });
```
EdwinGuzman: Why are you using the error to update the state for `saved`?  I guess if there is no error, then it successfully saved? Does the server return a success or object you can use instead?

src/components/PlsIDForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { State } from "../reducers/index";
+import { LibraryData } from "../interfaces";
+import { CheckSoloIcon } from "@nypl/dgx-svg-icons";
+import Input from "./reusables/Input";
+
+export interface PlsIDFormOwnProps {
+  uuid: string;
+  store: Store<State>;
+  fetchLibrary: (uuid: string) => LibraryData;
+  current?: string;
```
EdwinGuzman: `currentId` ?

EdwinGuzman:
I think it’d be good to have an explanation of what the pls id is for users and for developers. So a short description before the form explaining what it is and what it will be used for when an admin enters a value, and for devs comments to explain what the value is for.

src/components/LibrariesListItem.tsx
adriana-alter: I do think it's important enough to view in the header, unless we implement a way to search for it (which would be a separate ticket).  But I see what you mean about it being confusing; I'll make it clearer.

src/components/reusables/Input.tsx
adriana-alter: Yes, yes it is :(

src/components/PlsIDForm.tsx
adriana-alter: Exactly: if there's no error, that means it saved.  The server does return something, but, given that we don't actually need to do anything with that value, I think it makes more sense to just do it like this.  We're already using this pattern in `EmailValidationForm` too.

adriana-alter:
Good idea about having a description/explanation; coming up in the next commit!

src/components/PlsIDForm.tsx
EdwinGuzman: Yeah, but it'll be better if you use the good response for this instead of relying on the empty error. I think I had reservations about the `EmailValidationForm`, too.

src/components/LibrariesListItem.tsx
EdwinGuzman: I thought that id was used about once every six months? It doesn't seem that useful. I figured it'd go with the other key value library information pairs in a table.

adriana-alter: Ok.  I was thinking more because we're going to be inputting all of the PLS IDs, so it would be good to be able to see which libraries don't yet have one without having to open all the panels.  But...what would you think about putting something in the header to indicate if a library DOESN'T have one?  Or (in a different branch) putting in a filter (just on the front end) to show just the libraries that don't have one?

EdwinGuzman: The filter option sounds very good because you can extend it to filter for other library attributes.

adriana-alter:
Yes--go ahead and merge yours; I'll wait to publish.

PR #28, opened 2019-07-18
Adding tslint-react-a11y and updating LibraryStageItem.

EdwinGuzman:
Resolves [SIMPLY-2129](https://jira.nypl.org/browse/SIMPLY-2129).

Mostly updated one component and its `select` element. The troubling pattern I mentioned in [circ-web PR 250](https://github.com/NYPL-Simplified/circulation-web/pull/250) is also happening here. It's not necessarily bad but I can see it being troublesome for screenreaders. The parent component doesn't update the `LibraryStageItem`'s prop values until the form is submitted, so aria-selected doesn't really work the right way until you submit. The proper way for it to work would be to save the value in the state for each dropdown in the component, and update accordingly. I think that is a lot of overhead for this... but we can rethink proper patterns latter if it becomes a huge issue.


src/components/LibraryStageItem.tsx
```diff
@@ -12,21 +12,22 @@ export default class LibraryStageItem extends React.Component<LibraryStageItemPr
       "production": "success",
       "cancelled": "danger"
     };
+    const { label, value } = this.props;
```
adriana-alter: 👍

adriana-alter:
I think there was a compelling reason we used this pattern (although I don't remember what it was!).  I'll look into how much overhead it would be to change.

PR #29, opened 2019-07-25
Updating react-router to v5

EdwinGuzman:
... and making the necessary app and unit test changes.

I am currently making this a draft before I finalize it because I want your input. Can you run the app and let me know what you think of the changes. They pretty much are:
* The header is rendered all the time. When the user is not signed in, the logout button does not appear.
* No conditional rendering in the index component based on the `username` variable. Instead, render the whole app and route depending on if `username` is available or not. So, let react-router handle it instead of handling it through rendering.

I can go into the code with more comments depending on your thoughts. Some things though (if you decide to go into the comments):
* the `require(imgSrc)` call in the Header component gave me a lot of pain. It seems virtually impossible to successfully call that in unit tests because mocha doesn't understand `require` and we compile the code using Typescript, not webpack or some other tool. So, I made the whole app pass an `imgSrc` prop to make it easier to test; it just doesn't look that pleasant.
* this is actually not very single-page app-like. When you log in, there's a call to `reload()` the page instead of routing to the correct component, and therefore the conditional rendering. The same thing occurs for logging out. Since flask handles both cases, it kind of seems unnecessary to have routing at all. I'm okay with removing it but who knows if later on we will want to handle more pages in this app (and then it'll come in handy).


adriana-alter:
This is really nice!  (And good idea re: `imgSrc`.) . One minor adjustment: now that the header, which says "Library Registry Interface", is on the log-in page, it looks weird to have the form also say "Library Registry Interface".  How about changing the form title to something like "Log In" (and having the button just say "Submit")...or just not having a form title...or...?  Also, I think the margin should be bigger (maybe even vertically center the form)...but not a big deal.

EdwinGuzman:
Yeah, good point. Missed that the title was redundant.

src/__tests__/index-test.tsx
```diff
@@ -0,0 +1,39 @@
...
+const RegistryAdmin = require("../index");
+import { BrowserRouter } from "react-router-dom";
+import App from "../components/App";
+
+describe("RegistryAdmin", () => {
+  let renderSpy;
+
+  beforeEach(() => {
+    renderSpy = Sinon.spy(ReactDOM, "render");
+    new RegistryAdmin({ username: "", imgSrc: "" });
```
EdwinGuzman: The `imgSrc` is used for testing. The default value is the path to the logo in the repo but making it an empty string for tests makes mocha happy (it cannot call `require(''...")`).

src/components/App.tsx
```diff
@@ -1,14 +1,51 @@
...
+        <Header
+          text="Library Registry Interface"
+          alt="SimplyE"
+          imgSrc={this.props.imgSrc}
+          logOut="/admin/log_out"
+          loggedIn={!!username}
+        />
+        <Route
+          path="/admin"
+          render={() => (
```
EdwinGuzman: Let react-router handle conditionally loading the components. There's a way to redirect to `/admin` if any other (potentially bad) url was hit but I don't think it's needed.

src/components/__tests__/App-test.tsx
```diff
@@ -4,24 +4,55 @@ import * as React from "react";
...
+  let store;
+
   beforeEach(() => {
-    wrapper = Enzyme.shallow(
-      <App />
+    store = buildStore();
+    context = { store, username: "" };
+    wrapper = Enzyme.mount(
+      <ContextProvider {...context}>
+        <MemoryRouter initialEntries={["/admin"]}>
```
EdwinGuzman: `MemoryRouter` is useful for tests (and react native). It's creating a router and declaring the path that we want.

src/index.tsx
```diff
@@ -20,23 +17,14 @@ class RegistryAdmin {
...
-      ReactDOM.render(
-        <ContextProvider {...config}>
-          <LogInForm title="Library Registry Interface" store={store} />
-        </ContextProvider>,
-        document.getElementById("landing-page")
-      );
-    }
+    ReactDOM.render(
+      <ContextProvider {...config}>
+        <BrowserRouter>
```
EdwinGuzman: `BrowserRouter` is the new `Router`.

package.json
```diff
@@ -27,7 +27,7 @@
     "react-bootstrap": "0.32.4",
     "react-dom": "^16.8.6",
     "react-redux": "^7.1.0",
-    "react-router": "^3.2.0",
+    "react-router-dom": "^5.0.1",
```
EdwinGuzman: `react-router`? Nope. Now it's `react-router-dom` which has `react-router` as a dependency now.

PR #30, opened 2019-07-29
Filter by missing pls

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2155.  At the moment, it's only filtering by whether the library has the `pls_id` attribute, but we can (very easily) add whatever else down the road.  The Filter's default behavior is to only show items that do have the specified attribute, but by clicking the button you can flip it so that it shows items that DON'T have the attribute (e.g. libraries whose pls_id we still need to add!).  The filter is reusable.

Screen Shot 2019-07-29 at 12 43 49 PM

<hr />

Screen Shot 2019-07-29 at 12 37 29 PM

<hr />

Here's what it looks like if you have more than one attribute to filter by (just as an example):

Screen Shot 2019-07-29 at 12 45 55 PM


EdwinGuzman:
I'm a bit confused. Why introduce a different way of filtering than the qa/production filter toggle? I guess because you do not want to initially filter by those values when the app loads.
We discussed before about toggles and checkboxes and said how the qa/prod toggle is fine because we want to see the change automatically. Checkboxes should work with a submit button so you can make all your selections instead of having the list update every time you click on one. It will be jarring to see the list update every time you click on a checkbox when you have more filters.

It's a bit confusing to me to have the positive and negative selection "have"/"do not have" and then select the filter but it works. When there are more filters, what if you wanted to select libraries that do not have pls_id but that are validated? I don't think the way it works right now you'd be able to do that, right?

src/components/Filter.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export interface FilterState {
+  flip: boolean;
+}
+
+export default class Filter extends React.Component<FilterProps, FilterState> {
+  constructor(props: FilterProps) {
+    super(props);
+    this.flip = this.flip.bind(this);
+    this.state = { flip: this.props.initialFlip || false };
```
EdwinGuzman: You can set the default prop value to false.

EdwinGuzman:
Also, in the "do not have" state, the background color is red. We've been using red to signify and error or that something is "off" (like in the qa mode toggle). So it's confusing to have something that is on - "only show libraries which DO NOT have plds_id" - with a red background.

src/components/Filter.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  flip() {
+    this.setState({ flip: !this.state.flip });
+    this.props.flipFilter();
+  }
+
+  render(): JSX.Element {
+    let [flipped, unflipped] = this.props.buttonText || ["DO NOT have", "have"];
+    let filterButton = <Button className="inline inverted top-align" content={this.state.flip ? flipped : unflipped} callback={this.flip} />;
+    let title = <h3>{this.props.title || "Show items which"} {filterButton}</h3>;
```
EdwinGuzman: I don't think there's an h1 or h2 on the page and maybe this doesn't need to be a heading.

src/components/Filter.tsx
```diff
@@ -0,0 +1,63 @@
...
+    let isFlipped = this.state.flip ? " flipped" : "";
+    let filterInputs = filterNames.map((k) => {
+      return (
+        <Input
+          className={`filter-box${isActive(k)}${isFlipped}`}
+          type="checkbox"
+          key={k}
+          name={k}
+          value={this.props.filterKeys[k]}
+          label={k}
```
EdwinGuzman: It'd be nice to have a mapping of the names so it's not `pls_id` in the tab. Similar to what you did in the complaints select in circulation admin.

src/components/Filter.tsx
```diff
@@ -0,0 +1,63 @@
...
+          callback={(e) => this.applyFilter(e)}
+        />
+      );
+    });
+    return (
+      <div className="filters">
+        {title}
+        <hr/>
+        <ul>{filterInputs}</ul>
```
EdwinGuzman: This is a `ul` which is great, bu the `filterInputs` are an array of `div`s.

src/components/LibrariesPage.tsx
```diff
@@ -83,6 +96,36 @@ export class LibrariesPage extends React.Component<LibrariesPageProps, Libraries
...
+      if (Object.keys(dict).indexOf(attr) >= 0) {
+        return !!dict[attr];
+      }
+      let subDicts = Object.values(dict).filter(v => isDict(v));
+      return !!subDicts.filter((sd) => hasAttr(sd, attr)).length;
+    };
+
+    let attributes = Object.keys(this.state.filters.attributes).filter(attr => this.state.filters.attributes[attr]);
+    libraries && attributes.forEach(attr => {
+      libraries = libraries.filter(l => this.state.filters.flipped ? !hasAttr(l, attr) : hasAttr(l, attr));
```
EdwinGuzman: Can you make `hasAttr(l, attr)` into a variable so you don't call it twice?

EdwinGuzman: Hmm, actually, maybe it's not needed.

src/components/reusables/Input.tsx
```diff
@@ -25,6 +25,7 @@ export default class Input extends React.Component<InputProps, {}> {
           id={this.props.name}
           defaultValue={this.props.value}
           readOnly={false}
+          defaultChecked={this.props.type === "checkbox" && this.props.value}
```
EdwinGuzman: This is fine but it makes more sense to have a `checked` prop rather than use the `value` prop for a checkbox.

src/components/LibrariesPage.tsx
```diff
@@ -83,6 +96,36 @@ export class LibrariesPage extends React.Component<LibrariesPageProps, Libraries
...
+    updatedFilters[filter] = !this.state.filters.attributes[filter];
+    this.setState({ filters: {flipped: this.state.filters.flipped, attributes: updatedFilters} });
+  }
+
+  flipFilter() {
+    this.setState({ filters: {flipped: !this.state.filters.flipped, attributes: this.state.filters.attributes} });
+  }
+
+  filterByAttribute(libraries) {
+    let hasAttr = (dict, attr) => {
```
EdwinGuzman: This is a nice recursive function but there are no tests for this. Can you make this a separate function outside of `filterByAttribute` so you can test both `hasAttr` and `filterByAttribute`?

EdwinGuzman:
Looking good so far. Can you comment the functions and classes using jsdoc style?

adriana-alter:
How about just using toggles instead of checkboxes?  (I realize that doesn't solve the issue of what happens if you want libraries that do not have pls_id but are validated.  But that doesn't seem like it needs to be solved right now; that can be a separate iteration if/when the need arises.)

Holding off on jsdoc comments until the code is finished.

Screen Shot 2019-08-01 at 12 48 39 PM

src/components/__tests__/LibrariesPage-test.tsx
```diff
@@ -253,4 +253,25 @@ describe("LibrariesPage", () => {
...
+    expect(filteredLibraries[0]).to.equal(libraryWithPLS);
+  });
+
+  it("should check whether a library has an attribute", () => {
+    let libraryWithPLS = modifyLibrary(libraries[0], {"pls_id": "12345"}, "basic_info");
+    expect(wrapper.instance().hasAttr(libraryWithPLS, "pls_id")).to.be.true;
+    expect(wrapper.instance().hasAttr(libraries[1], "pls_id")).to.be.false;
+  });
+
+  it.only("should convert between an attribute's display name and key name", () => {
```
EdwinGuzman: `only` left here.

src/components/LibrariesPage.tsx
```diff
@@ -106,22 +110,23 @@ export class LibrariesPage extends React.Component<LibrariesPageProps, Libraries
...
-      }
-      if (Object.keys(dict).indexOf(attr) >= 0) {
-        return !!dict[attr];
-      }
-      let subDicts = Object.values(dict).filter(v => isDict(v));
-      return !!subDicts.filter((sd) => hasAttr(sd, attr)).length;
-    };
+  hasAttr(dict, attr) {
+    let isDict = (dict) => dict && typeof(dict) === "object" && !Array.isArray(dict);
+    if (!isDict) {
```
EdwinGuzman: How does this work if you're not invoking the function but just calling it?

src/components/Filter.tsx
```diff
@@ -35,21 +39,21 @@ export default class Filter extends React.Component<FilterProps, FilterState> {
...
     let filterButton = <Button className="inline inverted top-align" content={this.state.flip ? flipped : unflipped} callback={this.flip} />;
-    let title = <h3>{this.props.title || "Show items which"} {filterButton}</h3>;
+    let title = <p>{this.props.title || "Show items which"} {filterButton}</p>;
     let filterNames = Object.keys(this.props.filterKeys);
     let isActive = (k) => this.props.filterKeys[k] ? " active" : "";
     let isFlipped = this.state.flip ? " flipped" : "";
     let filterInputs = filterNames.map((k) => {
       return (
-        <Input
+        <li><Input
```
EdwinGuzman: The `li` should have the `key` attribute here instead of `Input`.

src/components/__tests__/LibrariesPage-test.tsx
```diff
@@ -253,4 +253,25 @@ describe("LibrariesPage", () => {
...
+  it("should filter libraries by attribute", () => {
+    let libraryWithPLS = modifyLibrary(libraries[0], {"pls_id": "12345"}, "basic_info");
+    wrapper.setProps({ libraries: { libraries: [libraryWithPLS, libraries[1]] } });
+    wrapper.setState({ filters: { flipped: false, attributes: { "pls_id": true }}});
+    let filteredLibraries = wrapper.instance().filterByAttribute(wrapper.prop("libraries").libraries);
+    expect(filteredLibraries.length).to.equal(1);
+    expect(filteredLibraries[0]).to.equal(libraryWithPLS);
+  });
+
+  it("should check whether a library has an attribute", () => {
```
EdwinGuzman: Since it's a recursive function, can you also give an example with one or two more levels?

src/components/Filter.tsx
```diff
@@ -41,27 +43,30 @@ export default class Filter extends React.Component<FilterProps, FilterState> {
...
-        <li><Input
-          className={`filter-box${isActive(k)}${isFlipped}`}
-          type="checkbox"
-          key={k}
-          name={k}
-          checked={this.props.filterKeys[k]}
-          label={k}
-          callback={(e) => this.applyFilter(e)}
-        /></li>
+        <li className="filter-box">
```
EdwinGuzman: Add a `key` attribute here too.

src/components/reusables/Toggle.tsx
```diff
@@ -38,7 +43,7 @@ export default class Toggle extends React.Component<ToggleOwnProps, ToggleState>
  toggle(e) {
    e.preventDefault();
    let status = !this.state.on;
-   this.props.onToggle(status);
+   this.props.onToggle(status, this.props.name);
```
EdwinGuzman: I'm surprised that the qa/prod library toggle doesn't complain because the function you pass in for `onToggle` only has the `status` prop but not the `name` prop. Maybe make this optional in the declaration to be explicit.

EdwinGuzman:
Yes, hold off on jsdoc style comments but at least do the `/** name of function or description */` above the class/function if you can.

I think using the toggle is better. I'm still a bit concerned about the have/do not have button and I think Willa can help give an opinion.

I would just say for now to put the filters below the search box since most users expect filters to go after searching for something. This is not the typical search app since searching and the filters are not dependent but it's a good pattern to use. I also didn't realize that it was hidden in a panel at first. Might it help if you add a margin around the panel so you can see that there's a button/heading along with the down arrow?

src/components/__tests__/LibrariesPage-test.tsx
adriana-alter: Given the current structure of `LibraryData`, there's nowhere else for it to go...

adriana-alter:
I'd like to get this in, and then revisit the have/do not have button conversation at such time as we're trying to implement the filter elsewhere; the point of this was originally just to create an easy way to tell which libraries we haven't yet input a PLS ID for (but then I got carried away making an entire filter component!), so we're already out of scope...

Adding a margin looks bad, but I made the panel a more attention-getting color; does that help?

EdwinGuzman:
Right, I think it'd be good to focus on small parts and not let PRs get out of hand. Yes, this color grabs the attention more than the previous color.

We can revisit this later but I would disagree that it would be when adding a new filter. I think just this one example of have/do not have along with the toggle is enough to make it something to test and see how this pattern works. The other discussion about having more filters and then their combinations can be worked on much later and is not an issue now. I would suggest creating a ticket and working with Willa on that, not urgent and I do know it's a low user-base.

src/stylesheets/filter.scss
```diff
@@ -1,3 +1,17 @@
+.libraries-page > .panel:last-of-type {
```
EdwinGuzman: This would make more sense if it went in a `libraries_page.scss` file or maybe target the `.filter`'s Panel parent?

src/components/__tests__/LibrariesPage-test.tsx
EdwinGuzman: Yep, that's very true. The function you created is a nice generic function so any test data should be used to test the function itself, and not just in the context of the component. That's what I was thinking. Maybe something like:
```
const testObj = { testlv1: { testlv2: { testlv3: { testlv4: "some value" }}}};
expect(wrapper.instance().hasAttr(testOb, "testlv3")).to.be.true;
expect(wrapper.instance().hasAttr(testOb, "testlv4")).to.be.true;
```

If the `LibraryData` structure changes, you can be sure your function will work. It could also be passed to some util file so any component can use it, but not necessary for now.

adriana-alter:
What in particular do you find confusing or unintuitive about the have/do not have button?  It seems really straightforward to me, so I'm having trouble even thinking what the alternative would be.

PR #32, opened 2019-08-13
Add LoadingIndicator to LibrariesList

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2026; shows LoadingIndicator (instead of inaccurate "There are no libraries in this registry" message) when the page first loads, and when the user first toggles QA mode on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v1.3.3
+#### Added
+- Rendered a loading indicator component while the list of libraries is being loaded.
+
 ### v1.3.2
 #### Updated
 - Changed the label on the search form to reflect new functionality.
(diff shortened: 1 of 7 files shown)


EdwinGuzman:
This is fine but I'd like to see some kind of time prop added to slow down the loading modal. The content loads/refreshes too fast and that flash can be bad for some users. Slowing or faking a loading indicator to users is a good pattern but something that can be added later.

adriana-alter:
I think that's only an issue because you're testing it in local; on production, the list of libraries is long enough that there will be a non-flashy loading time...  If you still think it's worth adding the time prop, what amount of time do you think is reasonable?

EdwinGuzman:
That is true. I think it's between 2-3 seconds at a minimum that should be used but not a big deal.

EdwinGuzman:
I do think it's worth it for accessibility as well, but now is not necessarily the time.

PR #33, opened 2019-08-16
Adding and using typedoc.

EdwinGuzman:
Resolves [SIMPLY-2162](https://jira.nypl.org/browse/SIMPLY-2162).

Adding Typedoc to generate doc and updating Travis to deploy the static site on Github Pages.

[GH Pages](https://nypl-simplified.github.io/registry_admin/)


src/components/Filter.tsx
```diff
@@ -21,6 +17,9 @@ export interface FilterState {
   flip: boolean;
 }
 
+/**
+* A reusable class for generating a panel populated with toggles for each of a list of specified attributes
```
EdwinGuzman: Comments should be next to the class/function they describe.

src/components/PlsIDForm.tsx
```diff
@@ -30,10 +30,12 @@ export interface PlsIDState {
   saved: boolean;
 }
 
+/**
+ * The PLS ID is generated by the annual IMLS Public Libraries Survey.
+ * We use it to gather metric data such as "number of covered branches",
+ * which gives us a good measure of adoption of SimplyE.
+ */
 export class PlsIDForm extends React.Component<PlsIDFormProps, PlsIDState> {
-  // The PLS ID is generated by the annual IMLS Public Libraries Survey.  We use it to gather metric data
```
EdwinGuzman: This is more of Python's docstring style.

src/components/LibrariesPage.tsx
```diff
@@ -102,8 +102,7 @@ export class LibrariesPage extends React.Component<LibrariesPageProps, Libraries
   }
 
   /**
-  * setFilter()
-  * @param {string} filter - the library attribute to be filtered by
+  * @param {string} filter The library attribute to be filtered by
```
EdwinGuzman: Technically, `{type} nameOfParam` is not needed here because `Typedoc` takes care of figuring out the name and type of the param, but it doesn't hurt to have.

PR #34, opened 2019-08-21
Updating the Panel and Header components.

EdwinGuzman:
Resolves [SIMPLY-2229](https://jira.nypl.org/browse/SIMPLY-2229).

Updates the Header component from the Reusable Components package: https://github.com/NYPL-Simplified/reusable-components/pull/27
Also, adds on https://github.com/NYPL-Simplified/reusable-components/pull/25 for adding IDs to Panels.


src/components/LibrariesListItem.tsx
```diff
@@ -55,7 +55,9 @@ export default class LibrariesListItem extends React.Component<LibrariesListItem
     let style = this.state.color;
     let { name } = this.props.library.basic_info;
     return(
-      <li><Panel style={style} headerText={name} content={this.content()} /></li>
+      <li>
+        <Panel id={name.replace(/ /g, "")} style={style} headerText={name} content={this.content()} />
```
adriana-alter: Is the ID exposed to users/screen readers?  If so, doing `replace` with dashes or underscores will be a lot easier to read.

src/components/SearchForm.tsx
```diff
@@ -29,6 +29,7 @@ export default class SearchForm extends React.Component<SearchFormOwnProps, Sear
   render(): JSX.Element {
     let input = (
       <Input
+        id="search-input"
         key="form-input"
```
adriana-alter: Better to also change the key to `search-input`, or is it not worth it?

src/components/LibrariesListItem.tsx
EdwinGuzman: No, IDs don't get exposed to users or screenreaders. I'll uses dashes.

src/components/SearchForm.tsx
EdwinGuzman: Yeah, for consistency. Not a big deal since it's just a key for react.

adriana-alter:
Ok!  Fix the verb tenses in the Changelog like you did for the other one, and this should be good to go.

EdwinGuzman:
Thanks, just need two PRs in `reusable-components` to be reviewed.

PR #35, opened 2019-08-21
Adding unique id as library name for each PlsIDForm component that ge…

EdwinGuzman:
Attempting to resolve [SIMPLY-2231](https://jira.nypl.org/browse/SIMPLY-2232).
This and more accessibility come up when using axe. This PR is just for the PlsIDForm and uses the library's name to create a unique id. That should be unique enough.

Screen Shot 2019-08-21 at 10 29 43 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 #### Added
 - Typedoc for generating code documentation.
 #### Updated
-- Updated the Panel components to take in new props and using the Header from Reusable Components.
-- Updating the PlsIDForm to use an Id prop to render unique Ids on the page.
+- Updated the Panel components to take in new props and used the Header from Reusable Components.
+- Updated the PlsIDForm to use the `uuid` prop as an ID to its `Input` component for accessibility purposes.
 
 ### v1.3.3
 #### Added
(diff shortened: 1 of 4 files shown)


adriana-alter:
Is the unique ID actually going to be exposed to users in any way (e.g. as a label that could potentially be read by screen readers)?  If not, then it might be a better idea to use the UUID instead of the library name.  The registry database doesn't actually have a uniqueness constraint on library names.  (The circ database does.  So it should almost definitely be fine either way.  But still.)

EdwinGuzman:
No, it won't be exposed for anyone, it's just for screenreaders. `uuid` is fine.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Typedoc for generating code documentation.
 #### Updated
 - Updated the Panel components to take in new props and using the Header from Reusable Components.
+- Updating the PlsIDForm to use an Id prop to render unique Ids on the page.
```
adriana-alter: Let's not phrase this as "render on the page," given that the ID is only visible to screen readers; maybe just, like, "accept a unique ID prop for accessibility purposes."  Also, for the sake of consistency with the rest of the changelog: capitalize "ID", and use past tense ("used the Header", "updated the PlsIDForm").

PR #38, opened 2019-09-05
Inporting the Input component and using it in different forms to correctly render a label element.

EdwinGuzman:
This goes with [reusable-component PR 30](https://github.com/NYPL-Simplified/reusable-components/pull/30) and resolves [SIMPLY-2276](https://jira.nypl.org/browse/SIMPLY-2276). The `Input` component was imported and implemented. Since a `label` element should always be rendered with an `input` element, the UI changed for a few forms.
Screen Shot 2019-09-05 at 3 22 51 PM
Screen Shot 2019-09-05 at 3 22 55 PM

The form's label and the input's label elements are different semantically and refer to their respective element. But it's very odd to see redundant text in both labels (such as the PLS id form). Any better suggestion for a form label is welcomed.


adriana-alter:
Can one of the labels be hidden--but still accessible to screen readers, obviously!--in cases in which it's weird to have duplicate text?

EdwinGuzman:
Yes... but I'm not sure that may be the right approach. You need the label that is associated with the input to be semantically correct and it could be hidden, but that label should be visible for users to click on it to go to the right input element (since you can click both the label and input).

I think the general label for the form is not needed (that should also be a legend in a fieldset) and that could be removed but may look visually off. In the PLS ID example, the first "PLS ID" label with the underline is what I think can be removed.

EdwinGuzman:
What do you think about having the full name for the form:
Screen Shot 2019-09-18 at 2 56 47 PM

Then the label for the ID is labeled with the input.

adriana-alter:
Yes, good idea!  I think "Public Libraries Survey ID", though.

PR #39, opened 2019-10-02
Aggregates

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2268

Note: I changed the way the interface fetches library data, in the interests of making this feature provide admins with aggregate data as smoothly and quickly as possible.  Before: only the production libraries would initially get loaded, and then there would be a second server request to get the QA libraries once you clicked the QA toggle.  Now: because the whole point of this feature is that it can display info about how many QA libraries there are, all the libraries get fetched at the same time.  (The functionality from the users' perspective hasn't changed; you still only see the production list until you click the QA toggle.) . Hard to tell in local whether this will cause noticeable performance issues; we'll find out...?

How this feature works:

You can see the same data in two formats: list or charts.  In the list view, you initially just have the numbers and percentages for each stage, nicely formatted.

Screen Shot 2019-10-02 at 12 54 39 PM

If you click on "Show Library Names":

Screen Shot 2019-10-02 at 12 55 11 PM

The "Copy" button lets you copy the data with the pretty formatting intact.  (It works as expected re: the expanded/hidden list of names--i.e. you can copy either just the numbers, or the numbers and the names.) . For use cases where it's more helpful to just have the data in plain text, without the pretty formatting, you have the option of viewing and copying an unformatted version:

Screen Shot 2019-10-02 at 12 57 26 PM

Screen Shot 2019-10-02 at 12 57 44 PM

The Charts tab gives you a pie chart and a bar chart.  Unfortunately, you can't copy those (apparently there are issues with programmatically copying SVGs, and finding a workaround turned out to be more trouble than it was worth), but you can of course screenshot them.  If it turns out that there is a strong need to be able to copy them, we'll revisit this.

Screen Shot 2019-10-02 at 12 59 36 PM

Any changes you make to a library are immediately applied--e.g. if you change a library's stage to "cancelled," that data will immediately be reflected in the list and charts.

### DIFF
--- a/src/components/AggregateList.tsx
+++ b/src/components/AggregateList.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { Button } from "library-simplified-reusable-components";
 import { LibraryData } from "../interfaces";
+import DropdownButton from "./DropdownButton";
 
 export interface AggregateListProps {
   data: {[key: string]: LibraryData[]};
@@ -8,21 +9,26 @@ export interface AggregateListProps {
 
 export interface AggregateListState {
   copied: boolean;
-  expanded: boolean;
   styled: boolean;
   showConfirm: boolean;
+  production: boolean;
+  testing: boolean;
+  cancelled: boolean;
 }
 
 export default class AggregateList extends React.Component<AggregateListProps, AggregateListState> {
   private statsRef = React.createRef<HTMLUListElement>();
+  STAGES = ["Production", "Testing", "Cancelled"];
 
   constructor(props: AggregateListProps) {
     super(props);
     this.state = {
       copied: false,
-      expanded: false,
       styled: true,
-      showConfirm: false
+      showConfirm: false,
+      production: false,
+      testing: false,
+      cancelled: false
     };
     this.copy = this.copy.bind(this);
     this.toggleFormatting = this.toggleFormatting.bind(this);
@@ -40,10 +46,14 @@ export default class AggregateList extends React.Component<AggregateListProps, A
         content={`${this.state.styled ? "Remove" : "Restore"} Formatting`}
         className={className}
       />,
-      <Button
-        key="expand"
-        callback={this.toggleExpanded}
-        content={`${this.state.expanded ? "Hide" : "Show"} Library Names`}
+      <DropdownButton
+        mainContent="Library Name Display"
+        callback={(e) => { this.toggleExpanded(e); }}
+        menuContent={
+          [`${this.STAGES.every(x => this.state[x.toLowerCase()]) ? "Hide" : "Show"} All`].concat(
+            this.STAGES.map(x => (this.state[x.toLowerCase()] ? "Hide " : "Show ") + x)
+          )
+        }
         className={className}
       />,
       <Button
@@ -91,8 +101,17 @@ export default class AggregateList extends React.Component<AggregateListProps, A
     this.setState({ showConfirm: false });
   }
 
-  toggleExpanded() {
-    this.setState({ expanded: !this.state.expanded });
+  toggleExpanded(e) {
+    let [verb, category] = e.target.textContent.toLowerCase().split(" ");
+    let newState = {};
+    let newValue = verb === "show";
+    if (category === "all") {
+      this.STAGES.map(x => newState[x.toLowerCase()] = newValue);
+    }
+    else {
+      newState[category] = newValue;
+    }
+    this.setState({...this.state, ...newState});
   }
 
   toggleFormatting() {
@@ -123,7 +142,7 @@ export default class AggregateList extends React.Component<AggregateListProps, A
           { this.makeCategoryBar(category) }
         </section>
         {
-          this.state.expanded &&
+          this.state[category] &&
           <ul className={`${hasStyles ? "inner-stats-list " : ""}`}>
             {this.makeLibraryNameList(category)}
           </ul>
--- a/src/components/AggregateList.tsx
+++ b/src/components/AggregateList.tsx
@@ -1,0 +1,134 @@
+import * as React from "react";
+import { Button } from "library-simplified-reusable-components";
+import { LibraryData } from "../interfaces";
+
+export interface AggregateListProps {
+  data: {[key: string]: LibraryData[]};
+}
+
+export interface AggregateListState {
+  copied: boolean;
+  expanded: boolean;
+  styled: boolean;
+  showConfirm: boolean;
+}
+
+export default class AggregateList extends React.Component<AggregateListProps, AggregateListState> {
+  private statsRef = React.createRef<HTMLUListElement>();
+
+  constructor(props: AggregateListProps) {
+    super(props);
+    this.state = {
+      copied: false,
+      expanded: false,
+      styled: true,
+      showConfirm: false
+    };
+    this.copy = this.copy.bind(this);
+    this.toggleFormatting = this.toggleFormatting.bind(this);
+    this.toggleExpanded = this.toggleExpanded.bind(this);
+    this.hideConfirm = this.hideConfirm.bind(this);
+    this.makeLi = this.makeLi.bind(this);
+  }
+
+  render(): JSX.Element {
+    let className = "inline squared inverted left-align";
+    let buttons = [
+      <Button
+        key="format"
+        callback={this.toggleFormatting}
+        content={`${this.state.styled ? "Remove" : "Restore"} Formatting`}
+        className={className}
+      />,
+      <Button
+        key="expand"
+        callback={this.toggleExpanded}
+        content={`${this.state.expanded ? "Hide" : "Show"} Library Names`}
+        className={className}
+      />,
+      <Button
+        key="copy"
+        callback={this.copy}
+        content={this.state.copied ? "Copy Data Again" : "Copy Data"}
+        className={className}
+      />
+    ];
+    return (
+      <div className="list-view">
+        <div className="buttons">
+          { buttons }
+          <span
+            className={`copy-confirmation ${this.state.showConfirm ? "visible" : ""}`}
+          >Copied to clipboard</span>
+        </div>
+        <ul
+          className="stats-list"
+          ref={this.statsRef}
+          contentEditable
+          suppressContentEditableWarning
+        >
+          { Object.keys(this.props.data).map(category => this.makeLi(category)) }
+        </ul>
+      </div>
+    );
+  }
+
+  copy() {
+    let copyArea = this.statsRef.current;
+    copyArea.focus();
+    document.execCommand("selectAll");
+    let copy = document.execCommand("copy");
+    if (copy) {
+      document.execCommand("copy");
+      window.getSelection().removeAllRanges();
+      copyArea.blur();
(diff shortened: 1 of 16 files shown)


EdwinGuzman:
When we went through it together, we saw that it will be good to round off the percentages so you don't get something like:
Screen Shot 2019-10-03 at 10 14 21 AM
Also, the "Copy" button should say something like "Copy text".
We also talked about moving this to a different page so that not everything is dumped on one page, but that would be something you and Risa talk about.

Yes, I do think there will be a performance boost on load because of how many libraries there are on the production server. Even now, it's about 2-3 seconds to load more libraries when you toggle the QA button.
In the case where you "show library names", maybe each of those panels should be collapsible in case you only want to see the libraries in "testing" or "cancelled"? Otherwise, you'd have to scroll a bit to get to the panel you want.

src/components/Stats.tsx
```diff
@@ -0,0 +1,199 @@
...
+}
+
+export interface StatsState {
+  expanded: boolean;
+  copied: boolean;
+  styled: boolean;
+}
+
+export default class Stats extends React.Component<StatsProps, StatsState> {
+  private statsRef = React.createRef<any>();
```
EdwinGuzman: I think you can do `HTMLUListElement` instead of `any`.

src/components/Stats.tsx
```diff
@@ -0,0 +1,199 @@
...
+    });
+    return sorted;
+  }
+
+  renderListView(sorted) {
+    let buttons = [
+      <Button
+        key="copy"
+        callback={this.copy}
+        content={this.state.copied ? "Copy Again" : "Copy"}
```
EdwinGuzman: The "Copy Again" is a good indicator to the user to let them know that their text was copied, but a more common pattern is a small notification saying "Copied to clipboard".

src/components/Stats.tsx
```diff
@@ -0,0 +1,199 @@
...
+          {this.makeLis(sorted)}
+        </ul>
+      </div>
+    );
+  }
+
+
+  renderChartView(sorted) {
+    const colors = {
+      "production": "#809F69",
```
EdwinGuzman: Where are these colors coming from? They should be the existing colors.

src/components/Stats.tsx
```diff
@@ -0,0 +1,199 @@
...
+              <Tooltip />
+              <Bar dataKey="value" barSize={40} />
+            </BarChart>
+        </ResponsiveContainer>
+      </div>
+    );
+  }
+
+  copy() {
+    let copyArea = (this.statsRef.current as any);
```
EdwinGuzman: You shouldn't need to do `as any` with a ReactRef.

src/components/Stats.tsx
```diff
@@ -0,0 +1,199 @@
...
+    return Object.keys(sorted).map((category) => {
+      return (
+        <li key={category} className={withStyles ? "stats-category" : ""}>
+          <section className={withStyles ? "stats-category-name" : ""}>
+            { categoryInfo(category) }
+          </section>
+          <ul className={`${withStyles ? "inner-stats-list " : ""}${this.state.expanded ? "" : "hidden"}`}>
+            {sorted[category].map((l) => {
+              return <li className="inner-stats-item" key={l.uuid}><p>{l.basic_info.name}</p></li>;
+            })}
```
EdwinGuzman: This is fine but you can do the function shortcut.

src/components/Stats.tsx
```diff
@@ -0,0 +1,199 @@
...
+      ];
+    };
+
+    return Object.keys(sorted).map((category) => {
+      return (
+        <li key={category} className={withStyles ? "stats-category" : ""}>
+          <section className={withStyles ? "stats-category-name" : ""}>
+            { categoryInfo(category) }
+          </section>
+          <ul className={`${withStyles ? "inner-stats-list " : ""}${this.state.expanded ? "" : "hidden"}`}>
```
EdwinGuzman: Maybe you only want to render the list of libraries if the state is expanded, instead of hiding it through css.

EdwinGuzman:
This looks good but I think you could organize the `Stats` component better if you take out the list view and the chart view (`this.renderListView()` and `this.renderChartView()`) into their own components. It's a big component doing a lot otherwise.

src/components/Stats.tsx
adriana-alter: They are the existing colors; they're the colors used for the panels in the library list.  But 1) those colors live inside ReusableComponents (because that's where Panel lives), and 2) those were generated by using SCSS to modify colors with variable names, so that doesn't work inside the `.tsx` file, so I had to figure out the resulting hex codes instead.

adriana-alter:
I'm stuck on one of the tests, but I wanted to get this back under review now anyway, since I'm out tomorrow.

There is now a `DropdownButton` component (maybe worth moving to ReusableComponents at some point) such that you can choose different combinations of which stages to show the library names for.

@killerpuppytails What do you think: keep the data feature at the top of the registry admin page (i.e. keep the interface to one page), or create a new page for it?

killerpuppytails:
> @killerpuppytails What do you think: keep the data feature at the top of the registry admin page (i.e. keep the interface to one page), or create a new page for it?

I vote top of the registry page, but then again, I'm rarely editing in this interface so whoever uses it more might not like content being pushed down. I'm fine with being outranked in that context.

adriana-alter:
I vote top of the registry page too.  It's in a collapsible panel, so it's not really pushing anything down significantly.

src/components/AggregateList.tsx
```diff
@@ -40,10 +46,14 @@ export default class AggregateList extends React.Component<AggregateListProps, A
...
       />,
-      <Button
-        key="expand"
-        callback={this.toggleExpanded}
-        content={`${this.state.expanded ? "Hide" : "Show"} Library Names`}
+      <DropdownButton
+        mainContent="Library Name Display"
+        callback={(e) => { this.toggleExpanded(e); }}
+        menuContent={
+          [`${this.STAGES.every(x => this.state[x.toLowerCase()]) ? "Hide" : "Show"} All`].concat(
```
EdwinGuzman: This would be nice to have in its own variable to see what is going on.

src/components/AggregateList.tsx
```diff
@@ -40,10 +46,14 @@ export default class AggregateList extends React.Component<AggregateListProps, A
...
+            this.STAGES.map(x => (this.state[x.toLowerCase()] ? "Hide " : "Show ") + x)
```
EdwinGuzman: And this can be a string interpolated value as well.

src/components/AggregateList.tsx
```diff
@@ -91,8 +101,17 @@ export default class AggregateList extends React.Component<AggregateListProps, A
...
-  toggleExpanded() {
-    this.setState({ expanded: !this.state.expanded });
+  toggleExpanded(e) {
+    let [verb, category] = e.target.textContent.toLowerCase().split(" ");
+    let newState = {};
+    let newValue = verb === "show";
+    if (category === "all") {
+      this.STAGES.map(x => newState[x.toLowerCase()] = newValue);
+    }
+    else {
```
EdwinGuzman: This should go on line 110.

src/components/DropdownButton.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from "react";
+import { Button } from "library-simplified-reusable-components";
+import { GenericWedgeIcon } from "@nypl/dgx-svg-icons";
+
+export interface DropdownButtonProps {
+  mainContent: string | JSX.Element;
+  menuContent: string[] | JSX.Element[];
+  callback: (event: React.MouseEvent) => void | any;
```
EdwinGuzman: Why return `any`?

src/components/DropdownButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+    this.renderMenu = this.renderMenu.bind(this);
+  }
+
+  render() {
+    return (
+      <div className={`btn dropdown-button-container ${this.props.className}`}>
+        <Button
+          content={[this.props.mainContent, <GenericWedgeIcon className={`${this.state.isOpen ? "up" : "down"}-icon`}/>]}
+          callback={this.toggle}
+          className={this.props.className + " dropdown-button-main"}
```
EdwinGuzman: String interpolation.

src/components/DropdownButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <Button
+          className={this.props.className}
+          callback={this.props.callback}
+          content={item}
+        />
+      );
+    });
+    return (
+      <ul className="dropdown-button-menu">
+        { (this.props.menuContent as any).map(item => {
```
EdwinGuzman: I think you can create an interface for `string[] | JSX.Element[]` although that may be too much (but better than using `any`).

src/components/AggregateList.tsx
```diff
@@ -0,0 +1,134 @@
...
+        content={`${this.state.styled ? "Remove" : "Restore"} Formatting`}
+        className={className}
+      />,
+      <Button
+        key="expand"
+        callback={this.toggleExpanded}
+        content={`${this.state.expanded ? "Hide" : "Show"} Library Names`}
+        className={className}
+      />,
+      <Button
```
EdwinGuzman: I think it's nice that this is the last button.

src/components/__tests__/AggregateList-test.tsx
```diff
@@ -0,0 +1,132 @@
...
+    let categories = wrapper.find(".stats-category");
+    expect(categories.length).to.equal(3);
+    ["Production", "Testing", "Cancelled"].forEach((name, idx) => {
+      let infoBar = categories.at(idx).find(".stats-category-name");
+      expect(infoBar.find("span").at(0).text()).to.equal(`${name}: ${getNumber(name, data)}`);
+      expect(infoBar.find("span").at(1).text()).to.equal(
+        ` (${(getNumber(name, data) / 4 * 100)}%)`
+      );
+    });
+    });
```
EdwinGuzman: Don't think this is indented correctly.

src/components/DropdownButton.tsx
adriana-alter: Trying to build in flexibility for if/when we want to reuse the component elsewhere...

adriana-alter:
With the circular focus, how do you then get out of the dropdown when you're ready to...?

Anyway, yes, seems like a different PR.

I made it so that the dropdown menu is always rendered (but can have a `hidden` class) because I couldn't get the `AggregateList` tests to work otherwise.

Yes, panel is supposed to be closed initially; I just set it to open for ease of development and then forgot to change it back.

Notification alignment fixed!
Screen Shot 2019-10-10 at 1 23 38 PM

EdwinGuzman:
Check nypl.org's header for examples. The main button's content that opens the menu turns into "close" and selecting it closes the menu. Yes, different PR.

It's looking good but either remove the mouse enter/leave events on the div or the button's toggle call because you can have the situation where you enter, click, and then exit and the reverse of what is expected will happen: the menu appears when you are not in the div and will disappear when you enter the div. (I vote the mouse events).

PR #40, opened 2019-11-01
Adobe data

adriana-alter:
Goes with https://github.com/NYPL-Simplified/library_registry/pull/142

Resolves https://jira.nypl.org/browse/SIMPLY-2359

Just like the registry stages data list: you can remove/restore formatting, and copy the data.  (The button for copying the data now lives in its own file, so it can be used in both places.  Easy to reuse, if we ever need it somewhere else.)

Screen Shot 2019-11-01 at 12 42 19 PM

@EdwinGuzman @killerpuppytails Are we still good with having the stats panel now that there are three tabs, or is it time to give it its own page?

### DIFF
--- a/src/__tests__/sharedFunctions-test.ts
+++ b/src/__tests__/sharedFunctions-test.ts
@@ -12,17 +12,12 @@ describe("getPercentage", () => {
   it("calculates the percentage out of a string containing a number", () => {
     expect(getPercentage(50, "200")).to.equal(25);
   });
-  it("calculates the percentage out of an array of numbers", () => {
+  it("calculates the percentage out of the sum of an array of numbers", () => {
     expect(getPercentage(50, outOf)).to.equal(25);
   });
-  it("calculates the percentage out of an array of strings containing numbers", () => {
+  it("calculates the percentage out of the sum of an array of strings containing numbers", () => {
     outOf = outOf.map(x => `${x}`);
     expect(getPercentage(50, outOf)).to.equal(25);
-  });
-  it("calculates the percentage out of an array of arrays", () => {
-    // The total is the combined lengths of the sub-arrays; 6, in this case.
-    outOf = [["a"], ["a", "b"], ["a", "b", "c"]];
-    expect(getPercentage(3, outOf)).to.equal(50);
   });
   it("optionally returns the percentage as a string", () => {
     expect(getPercentage(50, outOf, true)).to.equal("25%");
--- a/src/__tests__/sharedFunctions-test.ts
+++ b/src/__tests__/sharedFunctions-test.ts
@@ -1,0 +1,33 @@
+import { expect } from "chai";
+import { getPercentage } from "../utils/sharedFunctions";
+
+describe("getPercentage", () => {
+  let outOf;
+  beforeEach(() => {
+    outOf = [30, 55, 45, 70];
+  });
+  it("calculates the percentage out of a number", () => {
+    expect(getPercentage(50, 200)).to.equal(25);
+  });
+  it("calculates the percentage out of a string containing a number", () => {
+    expect(getPercentage(50, "200")).to.equal(25);
+  });
+  it("calculates the percentage out of an array of numbers", () => {
+    expect(getPercentage(50, outOf)).to.equal(25);
+  });
+  it("calculates the percentage out of an array of strings containing numbers", () => {
+    outOf = outOf.map(x => `${x}`);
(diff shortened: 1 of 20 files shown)


killerpuppytails:
Thanks @adriana-alter - it doesn't need its own page. Yet. :) (If we add one more data set, then I'll probably cave.)

src/__tests__/sharedFunctions-test.ts
```diff
@@ -0,0 +1,33 @@
...
+  beforeEach(() => {
+    outOf = [30, 55, 45, 70];
+  });
+  it("calculates the percentage out of a number", () => {
+    expect(getPercentage(50, 200)).to.equal(25);
+  });
+  it("calculates the percentage out of a string containing a number", () => {
+    expect(getPercentage(50, "200")).to.equal(25);
+  });
+  it("calculates the percentage out of an array of numbers", () => {
```
EdwinGuzman: Should include that it's the sum of the numbers in the array.

src/__tests__/sharedFunctions-test.ts
```diff
@@ -0,0 +1,33 @@
...
+    expect(getPercentage(50, outOf)).to.equal(25);
+  });
+  it("calculates the percentage out of an array of strings containing numbers", () => {
+    outOf = outOf.map(x => `${x}`);
+    expect(getPercentage(50, outOf)).to.equal(25);
+  });
+  it("calculates the percentage out of an array of arrays", () => {
+    // The total is the combined lengths of the sub-arrays; 6, in this case.
+    outOf = [["a"], ["a", "b"], ["a", "b", "c"]];
+    expect(getPercentage(3, outOf)).to.equal(50);
```
EdwinGuzman: I don't understand this use-case or where it would come in handy. From the above test, you're using the sum of the numbers in the array, but here you're using the length of arrays in the main array. It doesn't seem consistent.

src/utils/sharedFunctions.ts
```diff
@@ -0,0 +1,19 @@
+// Given a number x and a total number, calculate what percentage x is of the total.
+// The total can be passed directly in as a number or a string containing a number, or can
+// be derived by adding up 1) the values in an array of numbers, 2) the values in an array of strings containing numbers,
+// or 3) the lengths of the elements in an array of arrays.
+export function getPercentage(x: number, outOf: number | string | Array<string | number | Array<any>>, asString: boolean = false) {
```
EdwinGuzman: There is no test for the `asString` parameter.

EdwinGuzman:
I get this error when I have no patrons in my library. In the console, `library.basic_info.number_of_patrons` is `"0"` and it's a string. That's why I made a wrong comment about the `asString` parameter not being tested. I don't think you're expecting this as a string coming from the server when there are no users.

Screen Shot 2019-11-01 at 5 33 07 PM

src/__tests__/sharedFunctions-test.ts
adriana-alter: That's exactly how `AggregateList` is currently using it.  It passes in an array of three arrays (one for each production stage) containing library objects, and we need to know how many library objects there are in total.  But I'll try changing that for sake of consistency.

src/utils/sharedFunctions.ts
```diff
@@ -2,16 +2,15 @@
...
 // be derived by adding up 1) the values in an array of numbers, 2) the values in an array of strings containing numbers,
 // or 3) the lengths of the elements in an array of arrays.
-export function getPercentage(x: number, outOf: number | string | Array<string | number | Array<any>>, asString: boolean = false) {
+export function getPercentage(x: number, outOf: number | string | Array<string | number>, asString: boolean = false) {
   let total;
   if (Array.isArray(outOf)) {
-    total = outOf.map((el) => {
-      return Array.isArray(el) ? el.length : parseInt(el as string);
-    }).reduce((accum, next) => accum + next);
+    total = outOf.map((el) => { return parseInt(el as string); }).reduce((accum, next) => accum + next);
```
EdwinGuzman: Nothing serious but you can do `map((el) => parseInt(el as string))` to be consistent with the reduce callback function.

src/components/AdobeTab.tsx
```diff
@@ -0,0 +1,60 @@
+import * as React from "react";
```
EdwinGuzman: Small comment: this is named `AdobeTab` and makes sense but the charts or list components are not named with `Tab`.

PR #41, opened 2019-11-06
Adding react-axe and a script to run it.

EdwinGuzman:
Resolves [SIMPLY-2377](https://jira.nypl.org/browse/SIMPLY-2377). Run it using `npm run dev-test-axe` but it shouldn't be used for active development since it is resource intensive and gets slow.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -17,6 +17,14 @@ class RegistryAdmin {
     div.id = "landing-page";
     document.getElementsByTagName("body")[0].appendChild(div);
 
+    // `react-axe` should only run in development and testing mode.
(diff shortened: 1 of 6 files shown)


src/index.tsx
```diff
@@ -17,6 +17,14 @@ class RegistryAdmin {
     div.id = "landing-page";
     document.getElementsByTagName("body")[0].appendChild(div);
 
+    // `react-axe` should only run in development and testing mode.
+    // Running this is resource intensive and should only be used to test
+    // for accessibility and not while active development.
```
adriana-alter: "While actively develop_ing_"...?

PR #43, opened 2019-11-07
Fixing accessibility issues found by react-axe.

EdwinGuzman:
Resolves [SIMPLY-2384](https://jira.nypl.org/browse/SIMPLY-2384). Fixes some issues found. There are some color contrast issues that react-axe picks up but they are in transitional states for panel headings but I think it's false since some of those colors don't even exist. Maybe it picks up a shade of color _during_ a hover transition.

As an aside, `recharts` throws a lot of warnings in the console (which didn't before!). I fixed one about having "400" as width and height in the `ResponsiveContainer` component. The [documentation](http://recharts.org/en-US/api/ResponsiveContainer) says one of those props needs to be a percentage. Unfortunately, if one prop is a percentage, it doesn't seem to mount in tests so I made it so, only for those specific tests, it's a static width and height. This will be nice to look into but I didn't want to spend too much time on it, and you are more familiar with it so you may know what the better fix is. According to the warning, if you're using 400 for both height and width, then you may not even need `ResponsiveContainer`, which may be the case from the screenshots.
400 x 400:
Screen Shot 2019-11-07 at 4 11 05 PM

50% x 400:
Screen Shot 2019-11-07 at 4 10 16 PM

I also reduced some output from the `AggregateList` tests which was because both libraries in the tests had the same uuid. I had to do this:
```
let productionLibrary2 = modifyLibrary(productionLibrary1, { "name": "Production Library 2" });
productionLibrary2.uuid = "UUID2";
```
because `modifyLibrary` doesn't update a value that is a root value. Honestly, this is an issue with the test and not updating the test data correctly and shouldn't be a problem in real life.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -38,14 +38,16 @@ export default class App extends React.Component<AppProps, {}> {
           logOut="/admin/log_out"
           loggedIn={!!username}
         />
-        <Route
-          path="/admin"
-          render={() => (
-            username ?
-              <LibrariesListContainer /> :
-              <LogInForm store={store} />
-          )}
-        />
+        <main>
+          <Route
+            path="/admin"
+            render={() => (
+              username ?
+                <LibrariesListContainer /> :
+                <LogInForm store={store} />
+            )}
+          />
+        </main>
       </div>
     );
   }
--- a/src/components/Charts.tsx
+++ b/src/components/Charts.tsx
@@ -4,22 +4,30 @@ import { LibraryData } from "../interfaces";
 
 export interface ChartsProps {
   data: {[key: string]: LibraryData[]};
+  width?: string | number;
+  height?: string | number;
 }
 
 export default class Charts extends React.Component<ChartsProps, {}> {
+  static defaultProps = {
+    width: "50%",
+    height: 400
+  };
+
   render() {
     const colors = {
       "production": "#809F69",
       "testing": "#FFCD61",
       "cancelled": "#DA5D62"
     };
+    const { width, height } = this.props;
     let chartData = Object.keys(this.props.data).map((category) => {
       return { "name": category, "value": this.props.data[category].length };
     });
 
     return (
         <div className="chart-view">
-          <ResponsiveContainer width={400} height={400}>
+          <ResponsiveContainer width={width} height={height}>
             <PieChart ref="pie-chart-svg">
               <Legend
                 payload={ chartData.map(
@@ -40,7 +48,7 @@ export default class Charts extends React.Component<ChartsProps, {}> {
               <Tooltip />
             </PieChart>
           </ResponsiveContainer>
-          <ResponsiveContainer width={400} height={400}>
+          <ResponsiveContainer width={width} height={height}>
             <BarChart ref="bar-chart-svg" width={400} height={400} data={chartData}>
               <CartesianGrid strokeDasharray="5 5" />
               <XAxis dataKey="name" interval={0} />
--- a/src/components/LibraryStageItem.tsx
+++ b/src/components/LibraryStageItem.tsx
@@ -15,7 +15,7 @@ export default class LibraryStageItem extends React.Component<LibraryStageItemPr
     const { label, value } = this.props;
 
     let select = (
-      <select key={label} name={label} defaultValue={value}>
+      <select key={label} name={label} defaultValue={value} aria-label={`Select ${label}`}>
         <option value="testing" aria-selected={value === "testing"}>Testing</option>
         <option value="production" aria-selected={value === "production"}>Production</option>
         <option value="cancelled" aria-selected={value === "cancelled"}>Cancelled</option>
--- a/src/components/__tests__/AggregateList-test.tsx
+++ b/src/components/__tests__/AggregateList-test.tsx
@@ -12,6 +12,8 @@ describe("AggregateList", () => {
   let data;
   let productionLibrary1 = modifyLibrary(testLibrary1, { "name": "Production Library 1", "registry_stage": "production" });
   let productionLibrary2 = modifyLibrary(productionLibrary1, { "name": "Production Library 2" });
+  productionLibrary2.uuid = "UUID2";
+
   beforeEach(() => {
     data = {
       "production": [productionLibrary1, productionLibrary2],
--- a/src/components/__tests__/Charts-test.tsx
+++ b/src/components/__tests__/Charts-test.tsx
@@ -16,7 +16,7 @@ describe("Charts", () => {
       "testing": [testLibrary1],
       "cancelled": [testLibrary2]
     };
-    wrapper = Enzyme.mount(<Charts data={data} />);
+    wrapper = Enzyme.mount(<Charts data={data} width={400} />);
   });
   it("renders a pie chart", () => {
     let pieChart = wrapper.find(".chart-view").children().at(0).find(".recharts-wrapper");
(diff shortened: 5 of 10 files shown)


src/components/App.tsx
```diff
@@ -38,14 +38,16 @@ export default class App extends React.Component<AppProps, {}> {
...
         />
-        <Route
-          path="/admin"
-          render={() => (
-            username ?
-              <LibrariesListContainer /> :
-              <LogInForm store={store} />
-          )}
-        />
+        <main>
```
EdwinGuzman: There should be at least one `main` element on the page.

src/components/LibraryStageItem.tsx
```diff
@@ -15,7 +15,7 @@ export default class LibraryStageItem extends React.Component<LibraryStageItemPr
     const { label, value } = this.props;
 
     let select = (
-      <select key={label} name={label} defaultValue={value}>
+      <select key={label} name={label} defaultValue={value} aria-label={`Select ${label}`}>
```
EdwinGuzman: I think ideally we'd have an actual `label` element. I was thinking we could have used the legend for each form and use `aria-labelledby` for the select element. It's not in the same component and it complicates things. The easiest thing to make this pass is to add an `aria-label` (I also didn't know if you would've wanted to have an extra label element thrown into the UI).

src/components/__tests__/Charts-test.tsx
```diff
@@ -16,7 +16,7 @@ describe("Charts", () => {
       "testing": [testLibrary1],
       "cancelled": [testLibrary2]
     };
-    wrapper = Enzyme.mount(<Charts data={data} />);
+    wrapper = Enzyme.mount(<Charts data={data} width={400} />);
```
EdwinGuzman: Just to pass the tests so the component can render.

src/components/reusables/Toggle.tsx
```diff
@@ -33,7 +32,13 @@ export default class Toggle extends React.Component<ToggleOwnProps, ToggleState>
       <form className="toggle-container">
         { label && <label className={this.state.on ? "active" : ""}><span>{label}</span></label> }
         <div className={`toggle${this.state.on ? " toggle-on" : ""}`}>
-          <button tabIndex={0} className={`slider${this.state.on ? " slider-on" : ""}`} aria-pressed={this.state.on} onClick={this.toggle} />
+          <button
+            tabIndex={0}
+            className={`slider${this.state.on ? " slider-on" : ""}`}
+            aria-label="Toggle button"
```
EdwinGuzman: Buttons need text and these don't have it, which is fine, but should have an aria-label.

src/components/Charts.tsx
```diff
@@ -4,22 +4,30 @@ import { LibraryData } from "../interfaces";
...
 export interface ChartsProps {
   data: {[key: string]: LibraryData[]};
+  width?: string | number;
+  height?: string | number;
 }
 
 export default class Charts extends React.Component<ChartsProps, {}> {
+  static defaultProps = {
+    width: "50%",
+    height: 400
```
adriana-alter: I don't know whether it's because of this, but...now the labels on the bar chart get smushed together (as per your second screenshot).

src/components/LibraryStageItem.tsx
adriana-alter: It doesn't seem worth having an extra label element for this; it would be one thing if we were trying to use the component across more interfaces, but for now...

src/components/__tests__/AggregateList-test.tsx
```diff
@@ -12,6 +12,8 @@ describe("AggregateList", () => {
   let data;
   let productionLibrary1 = modifyLibrary(testLibrary1, { "name": "Production Library 1", "registry_stage": "production" });
   let productionLibrary2 = modifyLibrary(productionLibrary1, { "name": "Production Library 2" });
+  productionLibrary2.uuid = "UUID2";
+
```
adriana-alter: You can change the UUID in the `modifyLibrary` call; just pass it in as another key: value pair.

src/components/__tests__/Toggle-test.tsx
```diff
@@ -63,4 +63,10 @@ describe("Toggle", () => {
     expect(onToggle.args[1][0]).to.be.false;
     isOn(false);
   });
+
+  it("should have an aria-label", () => {
+    let slider = wrapper.find("button");
+
+    expect(slider.prop("aria-label")).to.equal("Toggle button");
+  });
```
adriana-alter: If we're doing this, then let's also check for aria-labels in the tests for the other components that use them (I know some of them already have, but I don't think all of them do)...

src/components/LibraryStageItem.tsx
EdwinGuzman: Yeah, it's not an extra label element, it's just the attribute. It's either the attribute or the label element and the attribute doesn't change the UI.

src/components/Charts.tsx
EdwinGuzman: Yes, but that's only if you try to squeeze the page down to a mobile view. Anything larger and the bars are evenly spaced. I don't think any one will really be viewing this on mobile anyway. So either the `ResponsiveContainer` can go since it doesn't add much and the props are incorrect, or deal with a squished mobile view.

src/components/LibraryStageItem.tsx
adriana-alter: No, I just meant re: what you were saying about not throwing an extra label element into the UI: good call!

src/components/__tests__/AggregateList-test.tsx
EdwinGuzman: `let productionLibrary2 = modifyLibrary(productionLibrary1, { "uuid": "uuid2", "name": "Production Library 2" });` ? I tried that. I mentioned that the root properties don't get updated in the object but the children objects do.

src/components/__tests__/Toggle-test.tsx
EdwinGuzman: Sounds good.

src/components/__tests__/AggregateList-test.tsx
adriana-alter: Eeeeeeek.  I will fix that.  That is not the intended behavior.

adriana-alter: Here you go! https://github.com/NYPL-Simplified/registry_admin/pull/44

src/components/__tests__/LibraryStageItem-test.tsx
```diff
@@ -40,6 +40,11 @@ describe("LibraryStageItem", () => {
     expect(options.at(1).html()).to.contain("Production");
     expect(options.at(2).html()).to.contain("Cancelled");
   });
+  it("should have an aria-label", () => {
+    let select = wrapper.find("select");
+
+    expect(select.prop("aria-label")).to.equal(`Select ${label}`);
```
adriana-alter: Unnecessary blank line (or is it just being rendered weirdly on Github?).  Looks good otherwise!

EdwinGuzman: That was intentional as I like a link after variables and before the tests but I can remove.

src/components/__tests__/AggregateList-test.tsx
EdwinGuzman: Thanks.

src/components/__tests__/LibraryStageItem-test.tsx
adriana-alter: I usually like that too, but it looks weird to me because the test is only two lines long.  Not a big deal either way, though.

EdwinGuzman:
Does this need anything else?

adriana-alter:
Sorry, thought I had hit "approve"!

PR #45, opened 2019-11-13
Don't initially show cancelled libraries

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2407.

1) Show the cancelled libraries only when you switch to QA mode, not mixed in with the initial list of production libraries.

2) Labeling the toggle "QA mode" doesn't make sense anymore (since it's not just showing testing libraries); change text to "Show All".

### DIFF
--- a/src/components/LibrariesPage.tsx
+++ b/src/components/LibrariesPage.tsx
@@ -41,7 +41,7 @@ export class LibrariesPage extends React.Component<LibrariesPageProps, Libraries
     this.search = this.search.bind(this);
     this.clear = this.clear.bind(this);
     this.toggleQA = this.toggleQA.bind(this);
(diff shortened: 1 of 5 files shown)


src/components/LibrariesPage.tsx
```diff
@@ -99,7 +99,8 @@ export class LibrariesPage extends React.Component<LibrariesPageProps, Libraries
   }
 
   filter(libraries) {
```
EdwinGuzman: Not necessary now but it'll be good to update this to be `filterProduction` so it's obvious from the function name that it's filtering for libraries in production, since that's the only filtering that's happening.

adriana-alter: Good call; will go ahead and do that now.

PR #46, opened 2019-11-15
Using opds-web-client to 0.3.0 and updating to Typescript 3.7.

EdwinGuzman:
Resolves [SIMPLY-2414](https://jira.nypl.org/browse/SIMPLY-2414).

- Updated Travis to use Node 10.
- Updated to `opds-web-client` to 0.3.0 and Typescript to 3.7.2.
- Fixed minor code complaints.


src/components/__tests__/LogInForm-test.tsx
```diff
@@ -57,15 +57,27 @@ describe("LogInForm", () => {
...
-    let extraCheckbox = <input type="checkbox" key="extraCheckbox" name="extraCheckbox" aria-checked={false} />;
-    wrapper.setProps({ extraFields: [extraText, extraCheckbox] });
+    let extraText = (
+      <input
+        type="text"
+        key="extraText"
+        name="extraText"
+        placeholder="some text"
+      />);
+    let extraText2 = (
```
EdwinGuzman: This is annoying but `tslint-react-a11y` has a bug where it thinks all inputs need to have a placeholder or default text but that's not true for checkboxes. So making this a text input for now.

Here's the issue: https://github.com/microsoft/tslint-microsoft-contrib/issues/749
`tslint-react-a11y` uses `tslint-microsoft-contrib` under the hood but maybe it hasn't been updated.

package.json
```diff
@@ -55,13 +56,13 @@
...
-    "typedoc": "^0.15.0",
-    "typescript": "2.9.1",
+    "ts-loader": "^6.2.1",
+    "tslint": "^5.20.1",
+    "tslint-react-a11y": "1.1.0",
+    "typedoc": "0.15.2",
+    "typescript": "3.7.2",
     "uglifyjs-webpack-plugin": "^2.1.2",
-    "url-loader": "^0.5.7",
+    "url-loader": "2.2.0",
```
EdwinGuzman: This was to fix a security issue.

package.json
```diff
@@ -55,13 +56,13 @@
...
     "style-loader": "^0.13.1",
-    "ts-loader": "^4.3.1",
-    "tslint": "^4.0.0",
-    "tslint-react-a11y": "^1.0.0",
-    "typedoc": "^0.15.0",
-    "typescript": "2.9.1",
+    "ts-loader": "^6.2.1",
+    "tslint": "^5.20.1",
+    "tslint-react-a11y": "1.1.0",
+    "typedoc": "0.15.2",
```
EdwinGuzman: This was to fix a security issue.

PR #47, opened 2019-11-19
Add geographic info to data panel

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2413.  In the List tab of the Aggregate Data panel, when the library names are being displayed, there is now an option to display each library's state (as in, the geographical state it's located in, not as in, whether it's in the production vs. testing state).  The "Show Geographic Info" button is disabled if none of the library names are visible, since there's nowhere to show the info.  If a library is associated with multiple locations, all the states are displayed, in a comma-separated, deduplicated list.  If we don't have information about the library's state, the string "state unknown" is displayed instead.


EdwinGuzman:
Looks good. The explanation for why the button is disabled works here for me as a developer, but if as a user it's odd that there's a button that is disabled and there's no indication of what I can do to make it clickable. I was thinking you could not disable it and if it was clicked it would trigger the "Show All" for the "Library Name Display" feature. Then it'd show the names and the geographic location. But now I think that maybe that button should appear when any library name is shown. I think that makes more sense.

Does "state unknown" work or is it more like the state wasn't configured yet?

src/components/AggregateList.tsx
```diff
@@ -100,7 +115,38 @@ export default class AggregateList extends React.Component<AggregateListProps, A
...
+    let allAreas: string[] = areas.focus.concat(areas.service);
+    allAreas.map((a: string) => {
+      // Each string is in the format "Zip code/city, (state abbreviation)".  We're just interested in the state abbreviation
+      // right now, so we get it by pulling out any two-letter sequence between parentheses.
+      // (If there's something longer than two letters between parentheses, it's the server-generated string "unknown",
+      // which we don't need to add individually to areaString; if everything was unknown, we'll handle it later.)
+      let regExp: RegExp = /\(([^)]{2})\)/;
+      let match: string[] = regExp.exec(a);
+      if (!match) {
+        return;
```
EdwinGuzman: An empty string should be returned.

EdwinGuzman: Do you have a test for this? It should've thrown an error since you are explicitly returning a string type.

EdwinGuzman: Oh, I see, this is not in the function but the loop. I think `forEach` would be better so you don't create a new array every time `map` is called (there are a lot of libraries in production and this will run a lot).

adriana-alter:
I think "state unknown" works.  It _is_ that the state wasn't configured yet...but (at least for now) the only way for the state to be configured is for the library people to have configured it in the circ web interface (i.e. it's not like the person using this feature can configure it)...so if the library hasn't configured it, then the registry doesn't have the information, so the information is unknown (from the perspective of the person using this feature).  Is there something different you would prefer?

PR #49, opened 2019-12-04
Updating the opds-web-client to 0.3.2 and cleaning up code.

EdwinGuzman:
Updating the `opds-web-client` to 0.3.2 and cleaning up some files with the intention to remove/update how context is used here. It turned into cleaning a few things up and removing `LibrariesListContainer` and using `LibrariesPage` directly. The `LibrariesListContainer` was only used to retrieve the `store` in context and pass it on but that could've been done at the higher level in `app.tsx` by passing it as a prop to `LibrariesPage`.


src/components/__tests__/ContextProvider-test.tsx
```diff
@@ -21,7 +22,7 @@ describe("ContextProvider", () => {
     let context = instance.getChildContext();
     expect(context.store.getState().libraries).to.be.ok;
     expect(context.store.getState().library).to.be.ok;
-    expect(context.pathFor).to.equal(instance.pathFor);
```
EdwinGuzman: So, `pathFor` is actually not used in this app. It's not in the context so it's undefined, and it's not in the instance so that is undefined too. That's why this test passed but it should not be checked. Instead, I passed down `username` which does get added to the context.

src/components/__tests__/LibrariesPage-test.tsx
```diff
@@ -100,18 +99,12 @@ describe("LibrariesPage", () => {
...
 
-  it("should search", async() => {
-    let form = wrapper.find(Form).at(0);
-    const libraryPageSearch = () => {
-      wrapper.setState({ showAll: false });
-      search();
-    };
-    let formSubmit = Sinon.stub(form.instance(), "submit").callsFake(libraryPageSearch);
-    let spyClear = Sinon.spy(wrapper.instance(), "clear");
-
```
EdwinGuzman: It seemed this didn't do anything. I think the search method and functionality could be better tested.

src/components/__tests__/LibrariesPage-test.tsx
```diff
@@ -209,7 +202,9 @@ describe("LibrariesPage", () => {
   });
 
   it("should filter", () => {
-    let libraryWithPLS = modifyLibrary(libraries[0], {"name": "PLS", "pls_id": "12345"}, "basic_info");
+    let libraryWithPLS = modifyLibrary(
+      libraries[0], {"name": "PLS", "pls_id": "12345", "uuid": "UUID4"}, "basic_info"
```
EdwinGuzman: Just added the `uuid` to remove a warning in the console.

PR #50, opened 2019-12-23
Sort by year

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2450

Another tab in the Stats panel: sorts the list of libraries by what year they registered.

@EdwinGuzman This is time-sensitive; Andrew would like to have access to this data before the end of the year (and I'm out on Thursday and Friday).  I'm sure there's plenty of room for refactoring (and/or finally moving all of the stats stuff to its own page!), but unless you spot something that's a HUGE problem from a functionality standpoint, better to just merge this as soon as possible so he can grab the list, and then I'll go back and clean up the code afterwards.

Screen Shot 2019-12-23 at 4 11 20 PM

PR #51, opened 2020-01-14
Refactor yearly data tab

adriana-alter:
Lots of changes!  Ticket lives [here](https://jira.nypl.org/browse/SIMPLY-2451) and original PR lives [here](https://github.com/NYPL-Simplified/registry_admin/pull/50).

Refactoring:

I tried to do an HOC (as per your suggestion in the original PR), but it turned out to be WAY more trouble than it was worth, because of needing to deal with the refs for the copy button.  Instead, I de-duplicated a lot of stuff by breaking out a `StatsInnerList` component--the list of library names sorted by production/testing/canceled which appears in the List tab and the Yearly Data tab.  Also created a reusable CSS class for the light blue bars used in Adobe Data and Yearly Data.

New features:

-- The Yearly Data tab now displays numbers and percentages for each year, and for the stages within each year.

-- The Yearly Data tab now implements the `DropdownButton` so that you can view one year at a time (I noticed there was a LOT of scrolling on production).


src/components/AdobeTab.tsx
```diff
@@ -42,9 +42,11 @@ export default class AdobeTab extends React.Component<AdobeTabProps, AdobeTabSta
             let name = library.basic_info.name;
             let numberOfPatrons = parseInt(library.basic_info.number_of_patrons);
               return (
-                <li key={name} className={hasStyles ? "adobe-data-li" : ""}>
-                  <span>{name}: {numberOfPatrons} </span>
-                  <span>({getPercentage(numberOfPatrons, this.props.data.map(l => l.basic_info.number_of_patrons), true)})</span>
+                <li key={name} className={hasStyles ? "adobe-li" : ""}>
+                  <section className={hasStyles ? "header-bar" : ""}>
+                    <span>{name}: {numberOfPatrons} </span>
```
EdwinGuzman: For someone not familiar with this UI, you'll see "library_name: 6". It makes more sense to include that it's 6 patrons. Not part of the PR but just noticed that when switching between tabs and, specifically, when looking at the yearly data tab, it says "2019: 7" but doesn't explain much more.

adriana-alter: That...is a very good point.

adriana-alter: (For the yearly data tab, it's not the number of patrons; it's the number of new libraries.  I guess it's REALLY confusing!  Will fix.)

EdwinGuzman: Right. In that case maybe it should be "libraried added" or something.

src/components/StatsInnerList.tsx
```diff
@@ -0,0 +1,86 @@
...
+        }
+      </li>
+    );
+  }
+
+  render() {
+    let list = this.props.data && Object.keys(this.props.data).map(category => {
+      return this.makeLi(category);
+    });
+    return <ul className="stats-inner-list">{list}</ul> || "null";
```
EdwinGuzman: Why the `"null"` case here? When will it ever get to that case?

src/components/StatsInnerList.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  getGeographicInfo(library: LibraryData): string {
```
EdwinGuzman: As mentioned offline, this seems to return digits instead of the state information.

src/components/StatsInnerList.tsx
```diff
@@ -0,0 +1,86 @@
...
+  styled: boolean;
+  stagesToShow?: {[ key: string ]: boolean};
+  showGeographicInfo?: boolean;
+}
+
+export default class StatsInnerList extends React.Component<StatsInnerListProps, {}> {
+  makeCategoryBar(category: string): JSX.Element[] {
+    let catName = category.replace(category[0], category[0].toUpperCase());
+    let length = this.props.data[category].length;
+    let allLengths = Object.values(this.props.data).map(x => (x as any).length);
```
EdwinGuzman: Not a big deal, but you're calculating this allLengths every time you make a new `li`. So maybe it could be done in `render` to do it once. The only drawback is that now you're passing in a value down two functions, and maybe the trade-off isn't bad.

src/components/YearlyDataTab.tsx
```diff
@@ -2,22 +2,33 @@ import * as React from "react";
...
 export default class YearlyDataTab extends React.Component<YearlyDataTabProps, YearlyDataTabState> {
   private dataRef = React.createRef<HTMLUListElement>();
   constructor(props: YearlyDataTabProps) {
     super(props);
-    this.state = { styled: true };
+    this.state = { styled: true, yearsToShow: {} };
     this.toggleFormatting = this.toggleFormatting.bind(this);
   }
+  async componentWillReceiveProps() {
+    await Object.values(this.props.data).some(x => x.length > 0);
```
EdwinGuzman: Object.values() isn't asynchronous, and I'm not sure what this is doing.

src/components/YearlyDataTab.tsx
```diff
@@ -26,26 +37,56 @@ export default class YearlyDataTab extends React.Component<YearlyDataTabProps, Y
...
-    let names = (category) => <ul className="yearly-library-list">{category.map(l => <li>{l}</li>)}</ul>;
     let sortedByYear = this.sortByYear(this.props.data);
-    let years = Object.keys(sortedByYear).map(y =>
-      <li key={y} className={this.state.styled ? "year-li" : ""}><p>{y}</p>{categories(sortedByYear[y])}</li>
-    );
+    let showOrHideAll: string = `${Object.keys(sortedByYear).every(year => this.state.yearsToShow[year]) ? "Hide" : "Show"} All`;
+    let showOrHideYear = (year) => `${this.state.yearsToShow[year] ? "Hide" : "Show"} ${year}`;
+    let total = Object.keys(sortedByYear).map(y => this.getYearlyTotal(sortedByYear[y]));
+    let years = Object.keys(sortedByYear).map((y) => {
+      let yearlyTotal = this.getYearlyTotal(sortedByYear[y]);
```
EdwinGuzman: All the yearly totals are in `total`. I think you can do `let yearlyTotal = total[index]` and in the map do `(y, index)`. Just to make one less call to `getYearlyTotal`.

src/components/DataTab.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from "react";
```
EdwinGuzman: It doesn't look like this is being used anywhere.

src/components/StatsInnerList.tsx
adriana-alter: It won't.  I just had it in for development and forgot to take it out.

src/components/StatsInnerList.tsx
adriana-alter: As also mentioned offline: I can't replicate this :(

src/components/YearlyDataTab.tsx
adriana-alter: No--getting all the values for `this.props.data` is asynchronous.  We're waiting for there to be actual libraries in the arrays, not just `{production: [], testing: [], cancelled: []}`.  This is how the dropdown knows what years we're dealing with, so it knows what to make the hide/show options for.  (This was the only way I could get that to work; I tried a bunch of things.)

adriana-alter:
It's the only thing I found that worked, unless I restructure the component altogether (and I think I would also have to restructure `StatsInnerList`) to accept a different kind of data as props.  Which seems very not worth it, given that this is just for populating the dropdown menu.  The `async/await` doesn't appear to be hurting anything; do you foresee specific problems with it, or is the issue just that it's weird?

EdwinGuzman:
It's weird that you have to do this and seems like a workaround, like not doing it the "react" way. It doesn't hurt anything (I think) and if you want to restructure it later on (taking a break and viewing it with new eyes) that's fine.

adriana-alter:
It is very much a workaround, but it's worth it.  Anyway, yes, I can restructure later!

PR #53, opened 2020-02-10
Add months to yearly data

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2556  (Andrew is currently working on the metrics for which this information is relevant, so it would be good to get it out the door pretty quickly if possible.)

1) Sorts the list of libraries in the Yearly Data tab by month.
2) Optionally displays each library's month.

Screen Shot 2020-02-10 at 3 32 52 PM

### DIFF
--- a/src/components/StatsInnerList.tsx
+++ b/src/components/StatsInnerList.tsx
@@ -7,6 +7,7 @@ export interface StatsInnerListProps {
   styled: boolean;
   stagesToShow?: {[ key: string ]: boolean};
   showGeographicInfo?: boolean;
+  showMonths?: boolean;
 }
 
 export default class StatsInnerList extends React.Component<StatsInnerListProps, {}> {
@@ -23,7 +24,7 @@ export default class StatsInnerList extends React.Component<StatsInnerListProps,
     let libraries = this.props.data[category];
     return libraries.map((l) => {
       return (
-        <li className="inner-stats-item" key={l.uuid}><p>{l.basic_info.name}{this.getGeographicInfo(l)}</p></li>
+        <li className="inner-stats-item" key={l.uuid}><p>{l.basic_info.name}{this.getMonth(l)}{this.getGeographicInfo(l)}</p></li>
       );
     });
   }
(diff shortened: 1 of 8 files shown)


src/components/YearlyDataTab.tsx
```diff
@@ -43,6 +45,12 @@ export default class YearlyDataTab extends React.Component<YearlyDataTabProps, Y
             sortedByYear[year][catName].push(library);
           }
         }
+        // Order by month, within each category.
```
EdwinGuzman: I'd move this outside of `sortByYear` to not be so coupled. It might be useful later on to have a `sortByMonth` function.

src/components/StatsInnerList.tsx
```diff
@@ -52,6 +53,14 @@ export default class StatsInnerList extends React.Component<StatsInnerListProps,
     return <ul className="stats-inner-list">{list}</ul>;
   }
 
+  getMonth(library: LibraryData): string {
```
EdwinGuzman: Thinking way ahead, this might be nice as a util function if it's ever needed outside this component.

src/components/YearlyDataTab.tsx
```diff
@@ -65,11 +73,22 @@ export default class YearlyDataTab extends React.Component<YearlyDataTabProps, Y
     }
     this.setState({...this.state, ...{ yearsToShow }});
   }
+  toggleMonths() {
+    this.setState({...this.state, ...{months: !this.state.months}});
```
EdwinGuzman: You can just do ` this.setState({months: !this.state.months});`, similar to `toggleFormatting`. `setState` should already be doing the merging of the entire state with the attribute you want to update.

EdwinGuzman: I think eventually having a `toggleStateVal(stateKey)` function would be helpful to reduce all the functions that only toggle a specific key in the sate.

EdwinGuzman:
It's possible for a user to be confused and think that the month they see is when it is in the specific category instead of when it was added. There's no explanation that those months in the parenthesis is when a library was added, so you can also think of it like "oh, so it was in October when this library was canceled". Yes, this UI is only for us and Andrew _already_ knows what to expect, but just pointing out that it could be clearer.

src/components/StatsInnerList.tsx
```diff
@@ -52,6 +53,14 @@ export default class StatsInnerList extends React.Component<StatsInnerListProps,
...
+    if (!this.props.showMonths) {
+      return "";
+    }
+    const months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"];
```
EdwinGuzman: You don't need this here anymore.

PR #55, opened 2020-03-05
Additional email validation forms

adriana-alter:
Goes with https://github.com/NYPL-Simplified/library_registry/tree/validate-emails-to-register (which will have its own PR as soon as I'm done with the tests).  Resolves https://jira.nypl.org/browse/SIMPLY-2605

Screen Shot 2020-03-05 at 4 40 40 PM
Screen Shot 2020-03-05 at 4 40 49 PM
Screen Shot 2020-03-05 at 4 41 33 PM
Screen Shot 2020-03-09 at 4 59 42 PM

### DIFF
--- a/src/components/EmailValidationForm.tsx
+++ b/src/components/EmailValidationForm.tsx
@@ -1,11 +1,10 @@
 import * as React from "react";
-import { Form, Button } from "library-simplified-reusable-components";
+import { Form } from "library-simplified-reusable-components";
 import { connect } from "react-redux";
 import ActionCreator from "../actions";
 import { FetchErrorData } from "opds-web-client/lib/interfaces";
 import { Store } from "redux";
 import { State } from "../reducers/index";
-import { LibraryData } from "../interfaces";
 import { CheckSoloIcon, XIcon } from "@nypl/dgx-svg-icons";
 
 export interface EmailValidationFormState {
@@ -37,7 +36,7 @@ export class EmailValidationForm extends React.Component<EmailValidationFormProp
     this.state = { validated: false };
   }
 
-  convert(email: string): string {
+  convertEmailTitle(email: string): string {
     return (email.includes("_") ? email.replace("_", " ") : email.replace(" ", "_"));
   }
 
@@ -64,8 +63,9 @@ export class EmailValidationForm extends React.Component<EmailValidationFormProp
   }
 
   render(): JSX.Element {
-    let emailString = this.convert(this.props.email);
+    let emailString = this.convertEmailTitle(this.props.email);
     let emailAddress = this.props.libraryInfo[this.props.email];
+    // libraryInfo has "contact_validated", "help_validated", and "copyright_validated" properties.
     let validated = this.props.libraryInfo[`${this.props.email.split("_")[0]}_validated`];
     let alreadyValidated = validated && (validated !== "Not validated");
     let successText = `Successfully validated ${emailAddress}`;
--- a/src/components/EmailValidationForm.tsx
+++ b/src/components/EmailValidationForm.tsx
@@ -1,71 +1,88 @@
 import * as React from "react";
-import { Form } from "library-simplified-reusable-components";
+import { Form, Button } from "library-simplified-reusable-components";
 import { connect } from "react-redux";
 import ActionCreator from "../actions";
 import { FetchErrorData } from "opds-web-client/lib/interfaces";
 import { Store } from "redux";
 import { State } from "../reducers/index";
 import { LibraryData } from "../interfaces";
-import { CheckSoloIcon } from "@nypl/dgx-svg-icons";
+import { CheckSoloIcon, XIcon } from "@nypl/dgx-svg-icons";
 
 export interface EmailValidationFormState {
-  sent: boolean;
+  validated: boolean;
 }
 
 export interface EmailValidationFormOwnProps {
-  library: LibraryData;
+  email: string;
+  libraryInfo: {};
+  uuid: string;
   store: Store<State>;
-  fetchLibrary: (uuid: string) => LibraryData;
-}
-
-export interface EmailValidationFormDispatchProps {
-  validateEmail: (data: FormData) => Promise<void>;
 }
 
 export interface EmailValidationFormStateProps {
   error?: FetchErrorData;
 }
 
+export interface EmailValidationFormDispatchProps {
+  validateEmail: (data: FormData) => Promise<void>;
+}
+
 export interface EmailValidationFormProps extends EmailValidationFormOwnProps, EmailValidationFormDispatchProps, EmailValidationFormStateProps {}
 
 export class EmailValidationForm extends React.Component<EmailValidationFormProps, EmailValidationFormState> {
-
   constructor(props) {
     super(props);
-    this.sendEmail = this.sendEmail.bind(this);
-    this.state = { sent: false };
+    this.validate = this.validate.bind(this);
+    this.renderInfoText = this.renderInfoText.bind(this);
+    this.state = { validated: false };
   }
 
-  async sendEmail(data: FormData): Promise<void> {
+  convert(email: string): string {
+    return (email.includes("_") ? email.replace("_", " ") : email.replace(" ", "_"));
+  }
+
+  renderTitle(emailString: string, emailAddress: string, alreadyValidated: boolean): JSX.Element[] {
+    let titleString = `${emailString}${emailAddress ? ": " + emailAddress : ""}`;
+    let icon = ((alreadyValidated || this.state.validated) && !this.props.error) ? <CheckSoloIcon /> : <XIcon />;
+    return [<span>{titleString}</span>, icon];
+  }
+
+  async validate(email: string, data: FormData): Promise<void> {
+    data.append("email", email);
     await this.props.validateEmail(data);
-    await this.props.fetchLibrary(this.props.library.uuid);
-    this.setState({ sent: !this.props.error });
+    this.setState({ validated: true });
+  }
+
+  renderInfoText(emailString: string, emailAddress: string, validated: string, alreadyValidated: boolean): string {
+    if (!emailAddress) {
+      return `No ${emailString} configured`;
+    } else if (!alreadyValidated) {
+      return "Not validated";
+    } else if (!this.state.validated && !this.props.error) {
+      return `Validated: ${validated}`;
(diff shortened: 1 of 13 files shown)


EdwinGuzman:
There are three forms and if they all have the same email, when you validate one form, the email gets validated and only that form tells you that it was updated. When you refresh the page, however, the timestamps for all the three forms are updated. This makes sense since it's the same email but it's a little bit odd since the other forms were explicitly validated.

I haven't been able to test this scenario but I don't expect it to break (and can you confirm it): if each form has a different email and one is validated, the timestamp for the other two forms shouldn't be updated.

src/components/EmailValidationForm.tsx
```diff
@@ -1,71 +1,88 @@
 import * as React from "react";
-import { Form } from "library-simplified-reusable-components";
+import { Form, Button } from "library-simplified-reusable-components";
```
EdwinGuzman: You're not using `Button`. Also, you can remove line 8, `LibraryData` isn't used.

src/components/EmailValidationForm.tsx
```diff
@@ -1,71 +1,88 @@
...
 
-    let buttonText = hasEmail ? "Validate email address" : "No email address configured";
-    let buttonContent = <span>{buttonText}{icon}</span>;
-    let infoText = (alreadyValidated && !this.state.sent && !this.props.error) ? "Already validated" : null;
-    let successText = `Successfully validated ${this.props.library.urls_and_contact.contact_email}`;
+  renderInfoText(emailString: string, emailAddress: string, validated: string, alreadyValidated: boolean): string {
+    if (!emailAddress) {
+      return `No ${emailString} configured`;
+    } else if (!alreadyValidated) {
+      return "Not validated";
```
EdwinGuzman: Do you have a screenshot of this to add to the PR just as an example?

src/components/EmailValidationForm.tsx
```diff
@@ -1,71 +1,88 @@
...
+    } else if (!this.state.validated && !this.props.error) {
+      return `Validated: ${validated}`;
+    }
+  }
 
+  render(): JSX.Element {
+    let emailString = this.convert(this.props.email);
+    let emailAddress = this.props.libraryInfo[this.props.email];
+    let validated = this.props.libraryInfo[`${this.props.email.split("_")[0]}_validated`];
```
EdwinGuzman: I'm guessing that the `libraryInfo` object has a property like `contact_validated` which is what you are getting here? Some documentation on that would be helpful since it's not easily readable or easy to tell why you're doing this.

src/components/EmailValidationForm.tsx
```diff
@@ -1,71 +1,88 @@
...
-    let buttonText = hasEmail ? "Validate email address" : "No email address configured";
-    let buttonContent = <span>{buttonText}{icon}</span>;
-    let infoText = (alreadyValidated && !this.state.sent && !this.props.error) ? "Already validated" : null;
-    let successText = `Successfully validated ${this.props.library.urls_and_contact.contact_email}`;
+  renderInfoText(emailString: string, emailAddress: string, validated: string, alreadyValidated: boolean): string {
+    if (!emailAddress) {
+      return `No ${emailString} configured`;
+    } else if (!alreadyValidated) {
+      return "Not validated";
+    } else if (!this.state.validated && !this.props.error) {
```
EdwinGuzman: Is this right? If it's not validated and there's no error, then display "validated"?

Also, what if there is an email, it's not already validated, and this last `else if` fails? There's no default return case.

EdwinGuzman: Hmmm this makes sense because if `Form`'s `infoText` is undefined, nothing gets rendered, but if you didn't already know that or if `Form` gets updated in the future, you may want to be explicit about what is returned so there's no error.

src/components/EmailValidationSection.tsx
```diff
@@ -0,0 +1,30 @@
...
+import EmailValidationForm from "./EmailValidationForm";
+
+export interface EmailValidationSectionProps {
+  library: LibraryData;
+  store: Store<State>;
+}
+
+export default class EmailValidationSection extends React.Component<EmailValidationSectionProps, {}> {
+  render(): JSX.Element {
+    let emailForms = ["contact", "help", "copyright"].map((x) =>
```
EdwinGuzman: A comment around the context for why these three email values are being validated would be nice for developers.

src/components/EmailValidationForm.tsx
```diff
@@ -1,71 +1,88 @@
...
+    this.validate = this.validate.bind(this);
+    this.renderInfoText = this.renderInfoText.bind(this);
+    this.state = { validated: false };
   }
 
-  async sendEmail(data: FormData): Promise<void> {
-    await this.props.validateEmail(data);
-    await this.props.fetchLibrary(this.props.library.uuid);
-    this.setState({ sent: !this.props.error });
+  convert(email: string): string {
```
EdwinGuzman: Why is this needed? It seems like it's manipulating emails from the server?

src/components/EmailValidationSection.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <li className="well">
+        <EmailValidationForm
+          store={this.props.store}
+          uuid={this.props.library.uuid}
+          email={`${x}_email`}
+          libraryInfo={this.props.library.urls_and_contact}
```
EdwinGuzman: I'm not sure if rendering one form three times with the same "hiddenName" and "hiddenValue" is a problem. Probably not but something I can look into later.

src/components/EmailValidationForm.tsx
adriana-alter: No, it's converting the name of the email, not the actual email address.  Like, turning "contact_email" into "contact email" and vice versa.

EdwinGuzman: I see. Maybe `convertEmailTitle` would be better or something like that?

src/components/EmailValidationForm.tsx
adriana-alter: Yes, that's right.  `this.state.validated` isn't whether the email has previously been validated, it's whether you've clicked "validate" just now.  So:
-- If it's never been validated, it displays "not validated".
-- If it has been validated at some point in the past, but not by you having clicked it just now, it displays "validated".
-- If it's validated because you just clicked it, i.e. if `this.state.validated` is true, it displays "successfully validated" and turns green (that's the `successText` being passed to the Form).

src/components/EmailValidationSection.tsx
adriana-alter: I'm not finding anything online that suggests it would be a problem.  I can see how there would be issues if all the forms were somehow being submitted together, but since they only get submitted one at a time...

src/components/EmailValidationForm.tsx
EdwinGuzman: So, if you just clicked to validate it, `this.state.validated` becomes true and the `successText` is shown, that I get. But this last `else if` is checking that `this.state.validated` is not true and there's no error. Is it assuming that it was therefore already validated at `validated` time stamp?

adriana-alter:
I haven't been able to test the they-each-have-different-emails scenario either; I'll deal with it on Monday!  (I don't see any way it would break, but...)

src/components/EmailValidationForm.tsx
adriana-alter: Yes--that's the only other possibility.  We've already covered "there's no email address" and "there's an email address but it's never been validated."

src/components/EmailValidationSection.tsx
EdwinGuzman: Yeah, since each is its own form, there shouldn't be a problem for individual submissions. But I'm not sure how screenreaders handle that (if they read the whole page and see the same value three times or read individual forms and see the value once). In any case, nothing to worry about at the moment.

adriana-alter:
Totally fine if they have different emails!

EdwinGuzman:
Cool. So the previous comment

> There are three forms and if they all have the same email, when you validate one form, the email gets validated and only that form tells you that it was updated. When you refresh the page, however, the timestamps for all the three forms are updated. This makes sense since it's the same email but it's a little bit odd since the other forms were explicitly validated.

is this how it's suppose to work?

What did the last commit fix? Also, just minor comments about writing comments and you're all set.

adriana-alter:
That is how it's supposed to work, if they all have the same email address.  I agree it's a little bit odd, but all the alternatives I can think of seem worse.  Do you have an idea?

The last commit is to fix my discovery that when they have different email addresses and you validate one of them, the outdated info text AND the success text get displayed, like so:
Screen Shot 2020-03-09 at 1 22 48 PM

adriana-alter:
I'm pretty sure I added comments to all the places that you pointed out should have one; did I miss something?  (Or did you just mean that you're going to write more comments now?  The caffeine never quite kicked in...)

src/components/EmailValidationForm.tsx
EdwinGuzman: @adriana-alter here's an unanswered question.

src/components/EmailValidationForm.tsx
EdwinGuzman: @adriana-alter here too.

src/components/EmailValidationSection.tsx
EdwinGuzman: @adriana-alter

EdwinGuzman:
☕️ ☕️ ☕️ 

I'm not sure how to approach that... but it's small enough that maybe it's not a big issue? When you refresh all the old values get updated and can cause confusion, but maybe it's really not a problem. I'll approve but did note where you could use some more comments.

src/components/EmailValidationForm.tsx
adriana-alter: Ah.  The answer was "no."  I'll generate one--hang on...

src/components/EmailValidationForm.tsx
adriana-alter: I did; github is just being weird.
Screen Shot 2020-03-09 at 5 01 56 PM

src/components/EmailValidationSection.tsx
adriana-alter: And again.  Github being weird.

Screen Shot 2020-03-09 at 5 05 52 PM

PR #61, opened 2020-12-01
Use contact_validated instead of timestamp, and update existing tests

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3326. See https://jira.nypl.org/browse/SIMPLY-3303 for more context. Basically, the registry has been calculating when libraries were added based on the timestamp property...but, as Andrew noticed, it turns out timestamp is just the date of the most recent update rather than the date of creation, and is therefore not reliably helpful data (since the goal is usually to generate a list of brand-new libraries). It turns out we don't actually have the date of creation recorded anywhere (AAAH!). As a rough proxy, I'm switching over to using the date when the contact email was validated, since the library generally can't go live until that happens; not perfect, but less inaccurate than the current system. (Also, it's the only other date information we have.) This branch updates all the places that were using timestamp to use contact_validated instead. If the email has not been validated, then we use the timestamp to estimate when the library was created (since obviously it can't have been created later than the date of the most recent update to it).


adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3326.  See https://jira.nypl.org/browse/SIMPLY-3303 for more context.  Basically, the registry has been calculating when libraries were added based on the `timestamp` property...but, as Andrew noticed, it turns out `timestamp` is just the date of the most recent update rather than the date of creation, and is therefore not reliably helpful data (since the goal is usually to generate a list of brand-new libraries).  It turns out we don't actually have the date of creation recorded anywhere (AAAH!).  As a rough proxy, I'm switching over to using the date when the contact email was validated, since the library generally can't go live until that happens; not perfect, but less inaccurate than the current system.  (Also, it's the only other date information we have.)  This branch updates all the places that were using `timestamp` to use `contact_validated` instead.  If the email has not been validated, then we use the timestamp to estimate when the library was created (since obviously it can't have been created later than the date of the most recent update to it).

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v1.4.10
+#### Updated
+-
```
EdwinGuzman, line 5: Forgot the update text.

package.json
```diff
@@ -20,51 +20,51 @@
     "build-docs": "node_modules/.bin/typedoc --tsconfig tsconfig.json"
   },
   "dependencies": {
-    "@nypl/dgx-svg-icons": "^0.3.9",
+    "@nypl/dgx-svg-icons": "^0.3.13",
```
EdwinGuzman, line 23: Are all these updates from this PR or from a different branch that just merged into this one?

src/components/StatsInnerList.tsx
```diff
@@ -53,11 +60,18 @@ export default class StatsInnerList extends React.Component<StatsInnerListProps,
     return <ul className="stats-inner-list">{list}</ul>;
   }
 
+  guessYear(library: LibraryData): string {
+    let year = library.basic_info.timestamp && library.basic_info.timestamp.match(/20\d+/)[0];
```
EdwinGuzman, line 64: Can this be shortened to `const year = library.basic_info?.timestamp.match(/20\d+/)[0];`?

src/components/YearlyDataTab.tsx
```diff
@@ -88,13 +89,13 @@ export default class YearlyDataTab extends React.Component<YearlyDataTabProps, Y
...
           <section className={this.state.styled ? "header-bar" : ""}>
-            <span>{y}: {yearlyTotal} librar{yearlyTotal !== 1 ? "ies" : "y"} added</span>
+            <span>{y}: {yearlyTotal} librar{yearlyTotal !== 1 ? "ies" : "y"} validated</span>
             <span>({getPercentage(yearlyTotal, total, true)})</span>
           </section>
           {
             this.state.yearsToShow[y] &&
             <section className={this.state.styled ? "list-holder" : ""}>
-            { <StatsInnerList data={sortedByYear[y]} styled={this.state.styled} showMonths={this.state.months} /> }
+            { <StatsInnerList data={sortedByYear[y]} hasYear={y !== "Unknown"} styled={this.state.styled} showMonths={this.state.months} /> }
```
EdwinGuzman, line 98: You do similar checks in this component and in `StatsInnerList`:
```
let validated = library.urls_and_contact.contact_validated;
let year = validated === "Not validated" ? ....
```
Since you're passing `hasYear` here, maybe you can pass the validation as well so you don't have to do it again in the child component. Not necessary to do but just a thought

src/components/__tests__/LibrariesPage-test.tsx
```diff
@@ -13,16 +13,23 @@ import Stats from "../Stats";
...
   let wrapper: Enzyme.CommonWrapper<{}, {}, {}>;
   let store;
 
   beforeEach(() => {
+    let lib1 = modifyLibrary(testLibrary1, {registry_stage: "production", library_stage: "production" });
+    let lib2 = modifyLibrary(testLibrary2, {registry_stage: "production", library_stage: "production" });
     libraries = [
-      modifyLibrary(testLibrary1, {registry_stage: "production", library_stage: "production"}),
-      modifyLibrary(testLibrary2, {registry_stage: "production", library_stage: "production"})
+      modifyLibrary(lib1, {"contact_validated": "Fri, 12 May 2019 17:07:40 GMT"}, "urls_and_contact"),
```
EdwinGuzman, line 31: Can you add multiple modifications to one `modifyLibrary` call?

src/components/__tests__/TestUtils.ts
```diff
@@ -86,6 +86,16 @@ export const modifyLibrary = (baseLibrary: LibraryData, newData: {[key: string]:
   return updatedLibrary;
 };
 
+export const validate = (baseLibrary: LibraryData, validatedAt?: string): LibraryData => {
```
EdwinGuzman, line 89: A `/**` style comment would be good to know what this is supposed to validate and how.

src/components/__tests__/TestUtils.ts
```diff
@@ -134,4 +144,16 @@ describe("TestUtils", () => {
...
   });
+
+  it("validates the contact email", () => {
+    let baseLibrary = testLibrary1;
+    expect(baseLibrary.urls_and_contact.contact_validated).to.be.undefined;
+    baseLibrary = validate(baseLibrary);
+    expect(baseLibrary.urls_and_contact.contact_validated).to.equal(baseLibrary.basic_info.timestamp);
+    baseLibrary = validate(baseLibrary, "Fri, 12 May 2018 15:05:34 GMT");
+    expect(baseLibrary.urls_and_contact.contact_validated).to.equal("Fri, 12 May 2018 15:05:34 GMT");
+    let noTimestamp = {...baseLibrary.basic_info, ...{timestamp: undefined}};
```
EdwinGuzman, line 155: You should be able to do `let noTimestamp = {...baseLibrary.basic_info, timestamp: undefined };` instead of having the second argument in a destructured object. Same for line 94 above.

src/components/__tests__/StatsInnerList-test.tsx
```diff
@@ -2,22 +2,24 @@ import { expect } from "chai";
...
 import * as React from "react";
-import { testLibrary1, testLibrary2, modifyLibrary } from "./TestUtils";
+import { testLibrary1, testLibrary2, modifyLibrary, validate } from "./TestUtils";
 import StatsInnerList from "../StatsInnerList";
 
 describe("StatsInnerList", () => {
   let wrapper;
   let data;
   let productionLibrary1 = modifyLibrary(testLibrary1, { "name": "Production Library 1", "registry_stage": "production" });
+  productionLibrary1 = validate(productionLibrary1, productionLibrary1.basic_info.timestamp);
```
EdwinGuzman, line 12: This didn't make sense to me at first but it does after checking out the `validate` function. Add a comment here though to make this process easier to understand.

EdwinGuzman:
This looks good but maybe call the `validate` function something else. It's more like modifying the library's created at timestamp or adding an "unknown" value. It's also confusing because it's not actually validating or modifying _real_ data - it's just for the tests.

src/components/YearlyDataTab.tsx
EdwinGuzman: After looking at all the files, there's the `validate` function from the TestUtils but maybe you can create a similar function that does something similar with real data rather than with the test data. That way you can update the library's object once instead of checking and updating values in two different components.

CHANGELOG.md
adriana-alter: 🙀

package.json
adriana-alter: Just from pulling the newest version of `master`.

src/components/StatsInnerList.tsx
adriana-alter: Sure!  (Putting the question mark after `timestamp`, though; there shouldn't be any way for `basic_info` to be missing.)

src/components/__tests__/LibrariesPage-test.tsx
adriana-alter: No...but you can use the `validate` function here.  I just forgot to go back and do it.

src/components/StatsInnerList.tsx
```diff
@@ -53,11 +60,19 @@ export default class StatsInnerList extends React.Component<StatsInnerListProps,
     return <ul className="stats-inner-list">{list}</ul>;
   }
 
+  guessYear(library: LibraryData): string {
+    const [year, formattedYear] = findYear(library.basic_info.timestamp, "No later than ", " No information available");
+    return formattedYear;
+    // return (year ? ` (No later than ${year})` : " (No information available)");
```
EdwinGuzman, line 66: Did you mean to keep this here?

EdwinGuzman (approved):
The update looks good just a minor comment about commented-out code. Sorry, I forgot to re-review this earlier

adriana-alter:
I did not mean to leave that line there; good catch!  (And no worries!)

PR #64, opened 2021-01-11
MonthlyDataTab

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3304.  Adds a Monthly Data tab to the aggregate data section; it has the same remove/restore formatting option and copy button as the other tabs do.  It defaults to displaying a list of the libraries validated during the current month and year, but has dropdowns so that you can opt to view libraries from any other month/year combination.

Screen Shot 2021-01-11 at 4 38 16 PM

### DIFF
--- a/src/components/MonthlyDataTab.tsx
+++ b/src/components/MonthlyDataTab.tsx
@@ -43,10 +43,9 @@ export default class MonthlyDataTab extends React.Component<MonthlyDataTabProps,
     let updatedTimes = {[key]: e.currentTarget.textContent, [valueToKeep]: this.state[valueToKeep]};
     let dataToShow = this.filter(updatedTimes.month, updatedTimes.year);
     this.setState({
-      ...updatedTimes, ...{
+      ...updatedTimes,
         dataToShow,
         styled: this.state.styled
-      }
     });
   }
 
--- a/src/components/MonthlyDataTab.tsx
+++ b/src/components/MonthlyDataTab.tsx
@@ -1,0 +1,139 @@
+import * as React from "react";
+import CopyButton from "./CopyButton";
+import { Button } from "library-simplified-reusable-components";
+import { LibraryData } from "../interfaces";
+import StatsInnerList from "./StatsInnerList";
+import DropdownButton from "./DropdownButton";
+import { months, hasLibraries, getMonth, toggleState, findYear } from "../utils/sharedFunctions";
+
+const today = new Date() as any;
+const thisMonth = today.toLocaleString("default", { month: "long" });
+const thisYear = findYear(today.toDateString())[0];
+export interface MonthlyDataTabProps {
+  data: {[key: string]: LibraryData[]};
+}
+
+export interface MonthlyDataTabState {
+  month: string;
+  year: string;
+  dataToShow: {[key: string]: LibraryData[]};
+  styled: boolean;
+}
+
+export default class MonthlyDataTab extends React.Component<MonthlyDataTabProps, MonthlyDataTabState> {
+  private monthlyDataRef = React.createRef<HTMLElement>();
+
+  constructor(props: MonthlyDataTabProps) {
+    super(props);
+    this.state = {
+      month: thisMonth,
+      year: thisYear,
+      styled: true,
+      dataToShow: this.filter(thisMonth, thisYear)
+    };
+  }
+
+  async componentWillReceiveProps() {
+    await hasLibraries(this.props.data);
+    this.setState({ dataToShow: this.filter(thisMonth, thisYear) });
+  }
+
+  updateTimeframe(e, key: string) {
+    let valueToKeep = ["month", "year"].find(x => x !== key);
+    let updatedTimes = {[key]: e.currentTarget.textContent, [valueToKeep]: this.state[valueToKeep]};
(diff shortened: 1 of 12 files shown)


src/stylesheets/stats.scss
```diff
@@ -52,34 +52,34 @@
...
-        margin-top: 5px;
-      }
-      .list-holder {
-        border-left: 1px solid $blue-dark;
-        border-right: 1px solid $blue-dark;
-        border-bottom: 1px solid $blue-dark;
-        padding: 10px 20px;
-      }
+    }
+    .header-bar {
```
EdwinGuzman, line 68: Why were these rules moved out?

src/utils/sharedFunctions.ts
```diff
@@ -34,3 +50,7 @@ export function findYear(date: string, resultText?: string, backupText?: string)
   let formattedYear = ` (${year ? (resultText || "") + year : backupText})`;
   return [year, formattedYear];
 }
+
+export function hasLibraries(data: {[key: string]: LibraryData[]}): boolean {
```
EdwinGuzman, line 54: 👌

src/components/__tests__/StatsInnerList-test.tsx
```diff
@@ -37,7 +37,7 @@ describe("StatsInnerList", () => {
     let newData = {...data, ...{"cancelled": []}};
     wrapper.setProps({ data: newData });
     let categories = wrapper.find(".stats-category");
-    ["Production", "Testing", "Cancelled"].forEach((name, idx) => {
+    ["Production", "Testing"].forEach((name, idx) => {
```
EdwinGuzman, line 40: Why the removal of "Cancelled" in the test?

src/components/__tests__/MontlyDataTab-test.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { LibraryData } from "../../interfaces";
+import { testLibrary1, modifyLibrary, validate } from "./TestUtils";
+import { findYear } from "../../utils/sharedFunctions";
+
+describe("MonthlyDataTab", () => {
+  let data: {[key: string]: LibraryData[]};
+  let wrapper: Enzyme.CommonWrapper<any, any, {}>;
+  const today = new Date() as any;
+  const thisMonth = today.toLocaleString("default", { month: "long" });
+  const thisYear = findYear(today.toDateString())[0];
```
EdwinGuzman, line 19: From the tests, this looks fine but I've found that using the actual result from the Date object can cause problems with time and timezones in Travis or other services. In general, if possible, it'll be good to mock the Date object value so it is always the same value.

src/components/MonthlyDataTab.tsx
```diff
@@ -0,0 +1,139 @@
...
+    await hasLibraries(this.props.data);
+    this.setState({ dataToShow: this.filter(thisMonth, thisYear) });
+  }
+
+  updateTimeframe(e, key: string) {
+    let valueToKeep = ["month", "year"].find(x => x !== key);
+    let updatedTimes = {[key]: e.currentTarget.textContent, [valueToKeep]: this.state[valueToKeep]};
+    let dataToShow = this.filter(updatedTimes.month, updatedTimes.year);
+    this.setState({
+      ...updatedTimes, ...{
```
EdwinGuzman, line 46: You don't need the second `...{`.

```
this.setState({
  ...updatedTimes,
  dataToShow,
  styled: this.state.styled
});
```
is fine.

EdwinGuzman (commented):
Looks great but just some questions.

src/components/__tests__/MontlyDataTab-test.tsx
adriana-alter: Sounds like a good idea, but I'm not clear on what you mean; do you mean hard-code a specific date (which I'd rather not do, since much of the point of the component is that, on load, it automatically shows you data for whatever the current month and year are...)?  Or...?

src/components/__tests__/StatsInnerList-test.tsx
adriana-alter: I changed the functionality slightly.  Old behavior: a `.stats-category` bar would be rendered for each of the three stages, even if there weren't any libraries under that category.  This looked kind of silly (and potentially confusing); I had been meaning to change it for a while, and this was a good opportunity.  New behavior: the `.stats-category` bar is only rendered if there are libraries to show.  TLDR: in `newData` in this test, the "cancelled" array is empty, so the interface shouldn't display a category bar that says "Cancelled."

src/stylesheets/stats.scss
adriana-alter: The `.header-bar` thing in `MonthlyDataTab` isn't inside a `ul`.

src/components/__tests__/MontlyDataTab-test.tsx
EdwinGuzman: Yes, hard code the date just for the tests. You can override the Date object with `sinon` and comment that for the purposes of the test the date is hard-coded to a specific date but in real life it would display the current month and year. There's no issue right now but I'd keep a tab on this in case it causes errors running tests in the future.

EdwinGuzman (approved):

PR #67, opened 2021-04-07
use v1.3.18 of reusable components

adriana-alter:


EdwinGuzman:
PR looks good but there are some test issues. Looks like some are related to dates? I wonder why they're breaking now.

PR #68, opened 2021-04-08
Date tests

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3684; fixing those mysteriously failing date-related unit tests...

I don't remember the rationale for mocking the Date object in these tests rather than just letting the component use today's date the way it normally would--do you?  It might be a better idea to just use the real date, and be able to test something closer to the component's actual behavior...


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## Changelog
 
+### v1.4.13
+#### Fixed
+- Fixed failing MonthlyDataTab tests.
+
+### v1.4.12
+#### Updated
```
EdwinGuzman, line 8: Since `1.4.12` wasn't published, why not bundle both changes together into `1.4.12` instead of `1.4.13`?

src/components/__tests__/App-test.tsx
```diff
@@ -41,7 +41,7 @@ describe("App", () => {
     expect(librariesPage.length).to.equal(0);
   });
 
-  it("should render the libraries page component", () => {
+  it.only("should render the libraries page component", () => {
```
EdwinGuzman, line 44: Remove `.only`.

EdwinGuzman:
The issue was that previously sometimes failed because the date that was being used kept changing.  Using the static dates fixed that but something else broke since then...

Try the real date again and see if that helps. Even if you can't fully test the dates, put the basic tests back. Better the trivial tests than nothing.

EdwinGuzman (approved):
Cool. Using the real date before wasn't much of an issue but I hope the tests don't break because of it later on for some weird date time issue.

PR #69, opened 2021-05-24
Comments

adriana-alter:
I went through the components and put in some basic comments to make things easier for the next person.


src/components/AggregateList.tsx
```diff
@@ -108,6 +111,7 @@ export default class AggregateList extends React.Component<AggregateListProps, A
   }
 
   toggleFormatting() {
+    // Status of the Remove/Restore Formatting button--determines whether CSS styles are applied to the data
```
EdwinGuzman, line 114: The comments are great, but comments like these that describe the complete function (rather than specific lines) or that describe the complete class should be written above the function or class declaration. If you have time, can you please convert the comments to go above each function? And for classes, start the comments with 

```
/*
 * Description goes here.
 */
class Something() {
...
```

webpack.common.js
```diff
@@ -49,7 +49,10 @@ var config = {
       },
       {
         test: /\.(png|woff|woff2|eot|ttf|svg).*$/,
-        loader: 'url-loader?limit=100000'
+        loader: 'url-loader?limit=100000',
+        options: {
+          esModule: false
```
EdwinGuzman, line 54: Why this update?

src/components/AggregateList.tsx
hokei: If an example is helpful, [this](https://github.com/NYPL-Simplified/circulation-web/blob/master/src/components/CirculationEventsDownloadForm.tsx) is probably a useful format/level of detail

adriana-alter:
Okay, I think I got them all!  (Running out of time, though!)

EdwinGuzman:
No, this is fine. Just remove the `esModule` config and there's a small typo that is making tests fail. After that, I'll merge it in but I won't publish a new version until I update the readme.

EdwinGuzman (approved):

PR #71, opened 2021-07-29
Simply 3852/decrease expensive ops

oliviawongnyc:
**Change**
- Adjusted all occurrences of the Panel component (in Filter, Stats, and LibrariesListItem) to only render contents on click.
- Edited existing tests, added new tests to account for this change.

**Reason**
- The parent components of the Panel render frequently, and rendering the Panel contents every time slows the app significantly.

**Tests**
- All 190 tests are passing locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Updated
 
 - Changed all instances of the Panel component to render contents only on click.
+- Upgraded to v1.3.19 of the reusable components package.
 
 ### v1.4.13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,218 +1,364 @@
 ## Changelog
 
+### v1.4.14
+
+#### Updated
+
+- Changed all instances of the Panel component to render contents only on click.
+
 ### v1.4.13
+
 #### Added
+
 - Put in comments for clarity.
 - Updated README for clearer startup instructions through Docker.
 
 ### v1.4.12
+
 #### Fixed
+
 - Fixed failing MonthlyDataTab tests.
+
 #### Updated
+
 - Upgraded to v1.3.18 of the reusable components package.
 
 ### v1.4.11
+
 #### Added
-- Added a Monthly Data tab to display a list of the libraries validated during any given month.
+
+- Added a Monthly Data tab to display a list of the libraries validated during
+  any given month.
 
 ### v1.4.10
+
 #### Updated
-- Modified the date display features to use the date when the library's contact email address was validated, rather than the timestamp property (which is the date when the library was last modified).
+
+- Modified the date display features to use the date when the library's contact
+  email address was validated, rather than the timestamp property (which is the
+  date when the library was last modified).
 
 ### v1.4.9
+
 #### Added
-- Added display for and ability to validate the help email and copyright email, along with the contact email.
+
+- Added display for and ability to validate the help email and copyright email,
+  along with the contact email.
 
 ### v1.4.8
+
 #### Added
+
 - Updated the `opds-web-client` package to 0.3.4.
 
 ### v1.4.7
+
 #### Added
-- Added an option for displaying the month each library was added, in the Yearly Data tab.
+
+- Added an option for displaying the month each library was added, in the Yearly
+  Data tab.
 
 ### v1.4.6
+
 #### Updated
+
 - Refactored and added additional features to the Yearly Data tab.
 
 ### v1.4.5
+
 #### Updated
+
 - Displayed the total number of Adobe IDs in the Adobe Data tab.
 
 ### v1.4.4
+
 #### Added
-- Added a Yearly Data tab to the Stats panel, displaying a list of the libraries sorted by what year they were added.
+
+- Added a Yearly Data tab to the Stats panel, displaying a list of the libraries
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,218 +1,364 @@
 ## Changelog
 
+### v1.4.14
+
+#### Updated
+
+- Changed all instances of the Panel component to render contents only on click.
+
```
EdwinGuzman, line 8: Similar to what's on line 34, add that the Reusable Components package was updated to 1.3.19.

src/components/Filter.tsx
```diff
@@ -26,38 +27,55 @@ export default class Filter extends React.Component<FilterProps, FilterState> {
   constructor(props: FilterProps) {
     super(props);
     this.flip = this.flip.bind(this);
-    this.state = { flip: this.props.initialFlip };
+    this.state = { flip: this.props.initialFlip, panelOpen: false };
```
EdwinGuzman, line 30: This should follow a similar pattern to the `flip` key and the `initialFlip` prop. So instead of `false`, it should be `this.props.panelOpen` and then set it to `false` by default.

src/components/Filter.tsx
```diff
@@ -26,38 +27,55 @@ export default class Filter extends React.Component<FilterProps, FilterState> {
...
-  * Toggle whether we're filtering via the specified attributes' presence vs. via their absence
-  */
+   * Toggle whether we're filtering via the specified attributes' presence vs. via their absence
+   */
   flip() {
     this.setState({ flip: !this.state.flip });
     this.props.flipFilter();
   }
 
+  handleFilterPanelClick() {
```
EdwinGuzman, line 53: Add a comment block with a short description of this function (similar to the other functions).

src/components/LibrariesListItem.tsx
```diff
@@ -12,43 +12,55 @@ export interface LibrariesListItemProps {
...
-export default class LibrariesListItem extends React.Component<LibrariesListItemProps, LibrariesListItemState> {
+export default class LibrariesListItem extends React.Component<
+  LibrariesListItemProps,
+  LibrariesListItemState
+> {
   constructor(props: LibrariesListItemProps) {
     super(props);
     const color = this.colorCode(Object.values(this.props.library.stages));
-    this.state = { color };
+    this.state = { color, panelOpen: false };
```
EdwinGuzman, line 26: You may want to set this to `this.props.panelOpen` instead of `false`, right? If you pass in `this.props.panelOpen=true` on the initial render, it won't be opened as expected in the initial render.

src/components/Stats.tsx
```diff
@@ -10,50 +10,63 @@ import MonthlyDataTab from "./MonthlyDataTab";
...
+export interface StatsState {
+  panelOpen: boolean;
+}
 
 /** The Aggregate Data panel. */
-export default class Stats extends React.Component<StatsProps, {}> {
+export default class Stats extends React.Component<StatsProps, StatsState> {
   constructor(props: StatsProps) {
     super(props);
+    this.state = { panelOpen: false };
```
EdwinGuzman, line 21: Same here as the other components, maybe this should be `this.props.panelOpen` instead of `false`.

src/components/__tests__/Stats-test.tsx
```diff
@@ -12,58 +12,113 @@ import YearlyDataTab from "../YearlyDataTab";
...
+        <Stats
+          libraries={[
+            testLibrary1,
+            testLibrary2,
+            productionLibrary1,
+            productionLibrary2,
+          ]}
+        />
+      );
+      let panelToggle = wrapper.find("button.panel-heading");
```
EdwinGuzman, line 60: New tests are good. Can you just write a comment here or above line 40 that explains why you need to first manually click on the panel to open it, such as "By default, the content of a Panel component is hidden. We need to manually trigger an onClick to open it to test the contents." or similar.

EdwinGuzman (commented):
This looks good but some minor comments.

src/components/Filter.tsx
EdwinGuzman: Actually, I take this back. This is the _state_ and not a _prop_. So this is fine to leave as is.

src/components/LibrariesListItem.tsx
EdwinGuzman: Same as before, this is the state and not the prop so this is okay to leave as is. The syntax confused me.

src/components/Stats.tsx
EdwinGuzman: Again, this is fine as is!

CHANGELOG.md
oliviawongnyc: Okay!

src/components/__tests__/Stats-test.tsx
oliviawongnyc: Added!

src/components/Filter.tsx
oliviawongnyc: Added!

EdwinGuzman (approved):
Looks good. You can update the code comment blocks to the following convention:
```
/**
 * description
 */
functionName()....
```

but it's not a blocker for approval.

PR #75, opened 2021-08-02
Simply 3854/search form UI changes

oliviawongnyc:
**Change**
- Adjusted SearchForm component to show loading message before results are returned.

**Reason**
- Before this change, SearchForm would return an error message even with valid search terms if the results hadn't yet come back from the server. This made for a confusing experience for the user.

**Tests**
- Adjusted SearchForm tests to account for this new behavior.
- All SearchForm tests are passing.

### DIFF
--- a/src/components/LibrariesPage.tsx
+++ b/src/components/LibrariesPage.tsx
@@ -31,6 +31,7 @@ export interface LibrariesPageState {
   searchTerm: string;
   qa: boolean;
   filters: { [key: string]: any };
+  receivedSearchResults: boolean;
 }
 
 export interface LibrariesPageProps
@@ -57,6 +58,7 @@ export class LibrariesPage extends React.Component<
       searchTerm: "",
       qa: false,
       filters: { flipped: false, attributes: { "PLS ID": false } },
+      receivedSearchResults: false,
     };
   }
 
@@ -75,6 +77,7 @@ export class LibrariesPage extends React.Component<
         inputName="name"
         clear={!this.state.showAll ? this.clear : null}
         resultsCount={libraries && libraries.length}
+        receivedSearchResults={this.state.receivedSearchResults}
       />
     );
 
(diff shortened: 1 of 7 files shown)


src/components/LibrariesPage.tsx
```diff
@@ -259,8 +262,14 @@ export class LibrariesPage extends React.Component<
       showAll: false,
       searchTerm: data.get("name") as string,
       filters: this.state.filters,
+      receivedSearchResults: false,
     });
-    await this.props.search(data);
+    try {
+      await this.props.search(data);
+    } catch (e) {
+      console.error(`No results found for '${this.state.searchTerm}'.`);
```
EdwinGuzman, line 270: Not receiving results from the API is technically not an error. It's possible a user search for a library that doesn't exist and the API returns 0 results. Instead of logging this as an error, it should be a warning `console.warn`.

But, the question that comes to mind is how useful this would be in the console since this message is meant for users. Maybe log the `e` error instead if that's more helpful to a dev.

src/components/SearchForm.tsx
```diff
@@ -90,11 +104,17 @@ export default class SearchForm extends React.Component<SearchFormOwnProps, Sear
...
+    } else if (this.props.resultsCount) {
+      let resultsNumber = `${this.props.resultsCount} ${
+        this.props.resultsCount > 1 ? "results" : "result"
+      }`;
+      message[
+        "success"
+      ] = `Displaying ${resultsNumber} for '${this.props.term}':`;
     } else {
-      message["error"] = `No results found for ${this.props.term}.`;
+      message["error"] = `No results found for '${this.props.term}'.`;
```
EdwinGuzman, line 117: Maybe a double quote is better in this case?

src/components/LibrariesPage.tsx
```diff
@@ -259,8 +262,14 @@ export class LibrariesPage extends React.Component<
       showAll: false,
       searchTerm: data.get("name") as string,
       filters: this.state.filters,
+      searchCompleted: false,
     });
-    await this.props.search(data);
+    try {
+      await this.props.search(data);
+    } catch (e) {
+      console.error("There was an error finding search results:", e);
```
EdwinGuzman, line 270: Update this to `.warn`.

EdwinGuzman (commented):
This is a great update. It's only missing one or two tests for the `LibrariesPage` component since new code was added there.

EdwinGuzman (approved):

PR #87, opened 2022-07-28
Builds login page, with fetch requests and unit tests

oliviawongnyc:
**Description**

- Builds login page
- Adds fetch requests to log_in/jwt and refresh_token endpoints
- Adds unit tests
- Adds comments
- Tidies up


src/components/LoginForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+        if (response.ok) {
+          return response.json();
+        }
+        if (response.status >= 400 && response.status < 502) {
+          setShowLoginError(true);
+        }
+        throw new Error('There was a problem authenticating this user.');
+      })
+      .then((data) => {
+        setAccessToken(data.access_token);
```
samanthaandrews, line 57: Do we also need to set the new refresh token value in the cookie? Or can the old refresh token be used more than once?

samanthaandrews (commented):
Just a couple questions. This looks really good! I do think the logic can get confusing about where we're storing access tokens vs refresh tokens but I think for this app this is a good enough pattern. I'll keep thinking about if there is benefit to only storing access and refresh tokens in the cookie vs in memory and type up some thoughts just to make sure I'm thinking about this clearly.

src/components/LoginForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const { setAccessToken } = useContext(TokenContext) as TokenContextValues;
+
+  const login = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    const body = new FormData();
+    body.append('username', username);
+    body.append('password', password);
+
+    fetch(process.env.QA_LOGIN as RequestInfo | URL, {
```
samanthaandrews, line 43: Hmm I don't see the `.env` file in these changes. Maybe it was added but not committed to git? Since it's not a secret, we can go ahead and commit this file.

samanthaandrews (approved):
🚀

PR #88, opened 2022-08-10
Fetches real library data

oliviawongnyc:
**Description**

- Adds fetch request to get real library data with user access token.
- Fixes/adds unit tests.
- Makes ActionBar separate component for legibility.

**Notes**

- The LibrariesList (the default view) is slow to load. This is because there are a lot of libraries and each accordion renders with all of its data even if the accordion is closed and the data is not visible. Edwin has given me the go-ahead to change the Accordion component in Reservoir to only render the content inside the panel if the Accordion is expanded. This should speed up the Registry Admin.
- There is a "maximum update exceeded" error in the console that appears whenever the window is resized. I have determined that this is an issue with the Tabs component inside the Accordion, and I believe is something that needs fixing on the Reservoir side (though there is a small chance it has to do with the React version we are using not being compatible with Reservoir). It doesn't crash the app, so I am not too worried at the moment, but definitely want to fix. There is a chance it may be alleviated with the change mentioned in the previous bullet. Edwin will be investigating as well.


src/components/__tests__/RegistryAdmin.test.tsx
```diff
@@ -88,40 +97,60 @@ describe('RegistryAdmin, with access token', () => {
 
     expect(libraryAccordions.length).toEqual(15);
   });
+
+  afterEach(() => {
+    jest.clearAllMocks();
```
samanthaandrews, line 102: I think we can enable this globally in the jest.config.js file by adding  `clearMocks: true` so that we don't have to add it to each test.

src/components/LibraryDetails.tsx
```diff
@@ -30,31 +26,51 @@ const LibraryDetails = ({ library }: LibraryDetailsProps) => {
...
+    stages.library_stage
+  );
+  const [registryStage, setRegistryStage] = useState<LibraryStage>(
+    stages.registry_stage
+  );
 
-  useEffect(() => {
-    setLibraryStage(stages.library_stage);
-  }, [stages.library_stage]);
+  // useEffect(() => {
```
samanthaandrews, line 36: Stray comments? Not sure if this is intentional or if we need it

src/components/__tests__/RegistryAdmin.test.tsx
```diff
@@ -88,40 +97,60 @@ describe('RegistryAdmin, with access token', () => {
...
-    expect(fetch).toHaveBeenCalledWith(
-      'https://qa-libraryregistry.librarysimplified.org/admin/refresh_token',
-      { headers: { Authorization: 'Bearer mockRefreshToken' }, method: 'POST' }
-    );
+    expect(fetch).toHaveBeenCalledWith(process.env.REFRESH, {
+      headers: { Authorization: 'Bearer mockRefreshToken' },
+      method: 'POST',
+    });
 
     Cookies.get = realGet;
```
samanthaandrews, line 126: Hm I don't quite understand this and maybe I missed it in a previous PR. Is it getting the cookies here for an assertion that we want to make?

src/components/LibrariesList.tsx
```diff
@@ -1,36 +1,56 @@
...
   isSimpleList: boolean;
+  libraries: LibraryData[];
 }
 
-const LibrariesList = ({ isSimpleList }: LibrariesListProps) => {
+const LibrariesList = ({ isSimpleList, libraries }: LibrariesListProps) => {
+  // isLoading is a flag used to render the SkeletonLoader while the app tries to
+  // refresh the accessToken.
+  const [isLoadingLibraries, setIsLoadingLibraries] = useState<boolean>(true);
   // Currently renders a list of mock library data.
```
samanthaandrews, line 21: Outdated comment

.env
```diff
@@ -1,3 +1,4 @@
 QA_DOMAIN=https://qa-libraryregistry.librarysimplified.org/admin
```
samanthaandrews, line 1: We should rename this to `REGISTRY_ADMIN_DOMAIN` since it will change based on the environment, qa or production. Then we should move `LOGIN`, `REFRESH`, and `FETCH_LIBRARIES` out into a `constants.ts` file. This is because I want to limit the number of env vars I have to define in the Dockerfile. Anything in the `.env` can be values that change depending on environment.

src/components/RegistryAdmin.tsx
```diff
@@ -1,110 +1,163 @@
...
+      // so the user is brought back to the LoginForm.
+      logout();
     }
   };
 
-  // If there is no accessToken in tokenContext, try refreshing. Else,
-  // setIsLoading to false. This will render the LibrariesList.
+  // Fetch the libraries with the accessToken.
+  const fetchLibraries = () => {
+    fetch(process.env.FETCH_LIBRARIES as RequestInfo | URL, {
```
samanthaandrews, line 107: I think there is an opportunity here to refactor some fetch requests into a utility (something like `authenticatedFetch` or `fetchWithHeaders`) that manages all of the refresh logic, since this logic will be the same for changing a library's status. I think we should do this in a separate PR though, perhaps the next one where we add the library status fetch requests. Just wanted to call it out here.

src/components/RegistryAdmin.tsx
```diff
@@ -1,110 +1,163 @@
...
+          return;
+        } else {
+          throw new Error('There was a problem fetching libraries.');
+        }
+      })
+      .then((data) => {
+        setLibraries(data.libraries);
+        setIsLoading(false);
+      })
+      .catch((err) => console.log(err.message));
```
samanthaandrews, line 126: Should we `setIsLoading(false)` here too and render an error component? This would be for a future ticket & PR. But want to get your thoughts on what you think might be a good way to display errors.

samanthaandrews (changes requested):
This looks great! So happy to see it coming together. Just a few things to cleanup/change.

src/components/LibraryDetails.tsx
oliviawongnyc: Was thinking I might use for the next PR, but will delete for now.

src/components/RegistryAdmin.tsx
oliviawongnyc: Sounds like a good idea. Okay!

src/components/RegistryAdmin.tsx
oliviawongnyc: Yeah, good callout. I have a ticket in the backlog regarding error handling.

src/components/__tests__/RegistryAdmin.test.tsx
samanthaandrews: Yes, thank you.

samanthaandrews (approved):
🚀

PR #90, opened 2022-08-17
Simply 4162/set up post requests

oliviawongnyc:
**Description**

- Adds POST requests to change library stages
- Updates Context to include libraries for easier manipulation of data
- Updates/adds tests


src/context/librariesContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+  setLibrariesInContext: Dispatch<React.SetStateAction<LibraryData[]>>;
+  setUpdatedLibrary: Dispatch<React.SetStateAction<LibraryData | undefined>>;
+};
+
+type LibrariesProviderProps = {
+  children: React.ReactNode;
+};
+
+export const LibrariesContext = createContext<LibrariesContextValues | null>(
+  null
```
samanthaandrews, line 15: Let's initialize these context values with `undefined` instead of `null` -- this will make some of the types easier to deal with.

Then, we can wrap this context inside of a function called `useLibraries` which will check for undefined and throw an error. Something like this can be added to the bottom of this file:

```
export default function useLibraries() {
  const context = React.useContext(LibrariesContext);
  if (typeof context === "undefined") {
    throw new Error("useLibraries must be used within a LibraryProvider");
  }
  return context;
}
```

Then, instead of passing the context name into the component, you can instead use:

```
const { librariesInContext } = useLibraries();
```

in the `LibrariesList.tsx` component for example.

We should use this same pattern with the token context too. This also eliminates the need to use the type coercion `as LibrariesContextValues` when you're using the context value.

samanthaandrews: This is the blog post where I discovered this pattern and it has served me well thus far: https://kentcdodds.com/blog/how-to-use-react-context-effectively

src/context/librariesContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+  useEffect(() => {
+    if (updatedLibrary) {
+      // Find the index of the library with the same uuid as the one
+      // that was recently updated.
+      const indexOfElementToReplace = librariesInContext.findIndex(
+        (library) => library.uuid === updatedLibrary.uuid
+      );
+
+      // Replace the library at that index with the updatedLibrary.
+      const updatedLibrariesInContext = librariesInContext.slice();
```
samanthaandrews, line 35: Interesting! I see what you're doing here. I'm not sure it's necessary to create a shallow copy of the libraries before updating the new library value. You could probably just splice() and then `setLibrariesInContext(librariesInContext)`, but I don't think this is bad at all! Very functional programming-esque.

src/context/librariesContext.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { createContext, Dispatch, useEffect, useState } from 'react';
+import { LibraryData } from '../components/RegistryAdmin';
+
+export type LibrariesContextValues = {
+  librariesInContext: LibraryData[];
+  setLibrariesInContext: Dispatch<React.SetStateAction<LibraryData[]>>;
```
samanthaandrews, line 6: yeah let's just make these `libraries` and `setLibraries`. It's clear through the rest of the context where they're coming from.

samanthaandrews (changes requested):
just a few small things otherwise looks good!

src/context/librariesContext.tsx
oliviawongnyc: Cool – this exactly describes my issue and the solution  😂. Thank you!

src/context/librariesContext.tsx
oliviawongnyc: I was attempting to not mutate the state 🤷‍♀️

samanthaandrews: Yeah I actually like it! Going to use in the future.

samanthaandrews (approved):
🚀

src/context/librariesContext.tsx
oliviawongnyc: Great!

PR #92, opened 2022-08-29
SIMPLY-4068: error handling

oliviawongnyc:
**Description**

- Adds error handling for fetch requests - feedback for devs in console and feedback for users in the UI
- Adds and updates tests where necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - SIMPLY-4069: Create functional login page.
 - SIMPLY-4067: Fetch real library data.
 - SIMPLY-4162: Make POST requests when a user changes a library's stage.
+- SIMPLY-4068: Add error handling for fetch requests.
 
 ## 1.4.15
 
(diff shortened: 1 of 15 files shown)


samanthaandrews (approved):
This is really great. Thanks for your patience as I took some time to review it!

My only suggestion would be to also implement a top-level React Error Boundary, mostly so that if a component error does pop up and Risa notices it, there would also be an accompanying stack trace that she could reference. Up to you if you'd prefer to add this error boundary in this PR or in a follow-up, doesn't matter to me.

oliviawongnyc:
> This is really great. Thanks for your patience as I took some time to review it!
> 
> My only suggestion would be to also implement a top-level React Error Boundary, mostly so that if a component error does pop up and Risa notices it, there would also be an accompanying stack trace that she could reference. Up to you if you'd prefer to add this error boundary in this PR or in a follow-up, doesn't matter to me.

Great! Thank you. I think I'll do the error boundary in another ticket/PR.

PR #95, opened 2022-09-20
SIMPLY-4194: add error boundary

oliviawongnyc:
- Adds library react-error-boundary, which eliminates the need to create a React Error Boundary class component and catches a wider variety of errors
- Adds fallback component, `ErrorFallback` which utilizes the DS `Notification` component to display the error message
- Adds tests


samanthaandrews (approved):
This is great. I think we should start using `react-error-boundary` in other projects now that I know it exists.

